PR #16, opened 2017-10-11
remove slow invariant

bradleyhd:
I've been working on a linter that disallows invocations of particular functions by name, and was testing it out on a ~5500 line source file. `getDescendantsWhere` consumes a very large amount of time (on the order of the number of lint errors and AST size).

I _think_ removing it would be okay as the `findWithParents` seems like enforces the same or a similar invariant. If not, maybe there's a cleaner way to enforce this? (i.e. not traversing from `$root` with every recursive call?)

Here's some XHPROF's from before and after!
before
after

In general, are you open to changes/alternative approaches in `find_offset_of_leading`?

### DIFF
--- a/src/find_offset_of_leading.php
+++ b/src/find_offset_of_leading.php
@@ -24,10 +24,6 @@ function find_offset_of_leading(
 
   $parents = null;
   $found = false;
-  invariant(
-    !C\is_empty($root->getDescendantsWhere(($it, $_p) ==> $it === $node)),
-    'node is not a descendant of root',
-  );
   $stack = $root->findWithParents($it ==> $it === $node);
   invariant(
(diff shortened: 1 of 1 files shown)


fredemmott:
This is great, thanks!

> In general, are you open to changes/alternative approaches in find_offset_of_leading?

Yes, but it would be good to get more test coverage first - I should have added them when I was working on it in the first place, sorry. It took a fair bit of iteration to actually get it to a point where I couldn't come up with anything that makes it spew out incorrect answers.

That said, if you're thinking of storing/caching it per-node, keep in mind that it needs to still give correct results on mutated ASTs - and as well as moves, this can include two different nodes having the same child node object.

PR #17, opened 2017-10-13
re-use stack while recursively calling find_offset_of_leading

bradleyhd:
Just looking for feedback here -- this is what I was envisioning for a follow up to #16. I'm assuming that an AST isn't able to be mutated while recursively calling `find_offset_of_leading` but ¯\\\_(ツ)\_/¯. I'm happy to write tests but frankly I don't know where to start or what to use as an example.

Thanks for indulging me here

Oh and here's some more XHPROFs for this one:
before
after

### DIFF
--- a/src/find_offset_of_leading.php
+++ b/src/find_offset_of_leading.php
@@ -17,6 +17,7 @@ use namespace HH\Lib\{C, Vec};
 function find_offset_of_leading(
   EditableNode $root,
   EditableNode $node,
+  ?vec<EditableNode> $stack = null
 ): int {
   if ($root === $node) {
     return 0;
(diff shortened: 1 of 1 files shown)


src/find_offset_of_leading.php
```diff
@@ -43,8 +47,11 @@ function find_offset_of_leading(
     $within_parent += $child->getWidth();
   }
 
+  array_pop($stack);
```
fredemmott: ```$stack = Vec\drop($stack, 1);```

bradleyhd:
Updated! I also ran a few linters and diff'd the output against master and everything seemed to be clean

fredemmott:
Thanks!

PR #18, opened 2017-10-20
added hhast-migrate to bin setting

holyshared:
Fix to be able to execute the following command.

```
vendor/bin/hhast-migrate
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,9 @@
 {
     "name": "hhvm/hhast",
-    "bin": [ "bin/hhast-lint" ],
+    "bin": [ 
+      "bin/hhast-lint",
+      "bin/hhast-migrate"
+    ],
     "require-dev": {
         "facebook/hack-codegen": "dev-master",
         "phpunit/phpunit": "^5.7",


fredemmott:
Thanks - this is currently intentional: running the existing migrations within the tree will break your code, so running hhast-migrate from a separate clone is recommended. Even if this weren't true, migrations are meant for one-off changes, so having it in vendor/ seems likely to be useful fairly rarely.

What's the use case you have for this?

As for why it will break your project: a partially migrated codebase (re shapes changes) can not be executed on any version of HHVM; if you're on the latest release, hhast isn't incompatible with 3.23, but the changes hhast-migrate makes require it. If you're on the next release, it will likely require 3.23, so you won't be able to run it from a codebase that's not already migrated.

holyshared:
I installed hhast and tried using the following command to change the code base from 3.21 to 3.22.  
However, because **vendor/bin/hhast-migrate** did not exist, it could not be executed.
This pull request is for that fix.

```
vendor/bin/hhast-migrate
```

I understood that it was intentional.
Thank you for answering.

> Thanks - this is currently intentional: running the existing migrations within the tree will break your code, so running hhast-migrate from a separate clone is recommended. Even if this weren't true, migrations are meant for one-off changes, so having it in vendor/ seems likely to be useful fairy rarely.

PR #20, opened 2017-10-25
use dict instead of switch in from_data codegen

bradleyhd:
For files with lots of tokens, the string `switch` in `editable_token_from_data.php` is a source of a fair amount of exclusive CPU time (bytecode dump suggests it walks each case and compares, TIL). An example:
before
before1

This is pretty much just a proof of concept, but using a `dict` of classnames seems to bring performance benefits (from what I assume is a hash table or something clever with a cheap lookup op). Same file after the changes:
after
after1

Not sure how this fairs in terms of readability, etc, but let me know!


codegen/editable_token_from_data.php
```diff
@@ -15,419 +15,213 @@ function editable_token_from_data(
...
-      return new HHAST\XHPStringLiteralToken($leading, $trailing, $token_text);
-    case 'XHP_body':
-      return new HHAST\XHPBodyToken($leading, $trailing, $token_text);
-    case 'XHP_comment':
-      return new HHAST\XHPCommentToken($leading, $trailing, $token_text);
-    case 'markup':
-      return new HHAST\MarkupToken($leading, $trailing, $token_text);
-    default:
-      throw new HHAST\UnsupportedTokenError($file, $offset, $token_kind);
+  static $class_map_with_text = dict[
```
fredemmott: It should be possible to have these as class constants now

codegen/editable_token_from_data.php
```diff
@@ -15,419 +15,213 @@ function editable_token_from_data(
...
+    '?->' => HHAST\QuestionMinusGreaterThanToken::class,
+    '...' => HHAST\DotDotDotToken::class,
+    '$$' => HHAST\DollarDollarToken::class,
+    '|>' => HHAST\BarGreaterThanToken::class,
+    '/>' => HHAST\SlashGreaterThanToken::class,
+    '</' => HHAST\LessThanSlashToken::class,
+    '<?' => HHAST\LessThanQuestionToken::class,
+    '?>' => HHAST\QuestionGreaterThanToken::class,
+  ];
+  $cls = idx($class_map, $token_kind);
```
fredemmott: `$class_map[$token_kind] ?? null`

codegen/editable_token_from_data.php
```diff
@@ -15,419 +15,213 @@ function editable_token_from_data(
...
+  if ($cls !== null) {
+    return new $cls($leading, $trailing);
   }
+  $cls = idx($class_map_with_text, $token_kind);
```
fredemmott: similar

src/__Private/codegen/CodegenEditableTokenFromData.php
```diff
@@ -23,6 +26,29 @@ public function generate(): void {
...
+      $kind = $token['token_kind'];
+      if (Str\lowercase($text) === Str\uppercase($text)) {
+        $class_map[$text] = sprintf('HHAST\\%sToken::class', $kind);
+      } else {
+        $class_map_with_text[$text] = sprintf('HHAST\\%sToken::class', $kind);
+      }
+    }
+
+    foreach ($tokens['variableText'] as $token) {
+      $kind = self::underscored($token['token_kind']);
```
fredemmott: whoops, I should move this to _Private\StrP like upper_camel

codegen/editable_token_from_data.php
bradleyhd: yields
```
codegen/editable_token_from_data.php:220:16,19: Was expecting a class, but got an untyped value (Typing[4026])
codegen/editable_token_from_data.php:224:16,19: Was expecting a class, but got an untyped value (Typing[4026])
```
(on `3.22.1-dev`)
:/

codegen/editable_token_from_data.php
bradleyhd: Do you mind elaborating on what that would look like? Our in-house style still avoids classes a lot, so I'm green in that department

fredemmott:
```Hack
class Foo {
  const BAR = dict['a' => 'b'];
}
```

via `$codegen_class->addConst($name, $value, $nullable_comment, $renderer)`

src/__Private/codegen/CodegenEditableTokenFromData.php
bradleyhd: I can include the move on the next PR update!

codegen/editable_token_from_data.php
fredemmott: :/ is this the full line that you have?

```
$cls = $class_map[$token_kind] ?? null;
```

bradleyhd: yessir, generated:
```
  $cls = $class_map[$token_kind] ?? null;
  if ($cls !== null) {
    return new $cls($leading, $trailing);
  }
  $cls = $class_map_with_text[$token_kind] ?? null;
  if ($cls !== null) {
    return new $cls($leading, $trailing, $token_text);
  }
```
`hh_client` doesn't like either of the `return` lines at `$cls`

fredemmott: okay - just a small nit, let's not do this change for now then.

codegen/editable_token_from_data.php
bradleyhd: I originally tried doing something like this instead of using `static`, but am not sure how to type it. For the class constant I'm required to add a type, but typing as `dict<string, string>` gives `Was expecting a class, but got a string (Typing[4026])` on the line with `new $cls`. Typing as `dict<string, classname>` needs a param for the classname. Not sure what's right here?

bradleyhd: doing
```
class TokenClass<T super HHAST\EditableNode> {
  const dict<string, classname<T>>
    MAP = dict[
      'abstract' => HHAST\AbstractToken::class,
      'and' => HHAST\AndToken::class,
      ...
    ];
```
makes `hh_client` complain with `Was expecting a class, but got an untyped value (Typing[4026])` just like the null coalesce 🤷‍♂️

codegen/editable_token_from_data.php
fredemmott: Actually, how is this valid? There's no common superclass defining a consistent constructor for these? That's likely why this is complaining

bradleyhd:
Pushed changes that satisfy the typechecker, but they feel kinda like cheating by adding a layer of obfuscation. I just saw your comment about `<<__ConsistentConstruct>>`, will give that a shot!

bradleyhd:
Hmm, I'm not sure I know how to make `<<__ConsistentConstruct>>` work in this case. Correct me if I'm wrong but this is the way I currently understand it:

All the codgen'd `XToken` classes inherit from `EditableToken`. `EditableToken`'s constructor takes 4 arguments. Some `XTokens` take 2 arguments, other 3 arguments, and fill in the rest with a combination of hard-coded values and default params. If I were to introduce two new intermediary classes, one for the 2-arg constructor variant and one for the 3-arg constructor variant, which both extend `EditableToken` and have `<<__ConsistentConstruct>>`, then they will not have enough information to construct a  `EditableToken` because the intermediary classes need to accept 4 args in order to pass them up to `EditableToken`, but by virtue of doing so the constructor signature that is expected of the grandchildren (`XToken`s) becomes a 4-arg variant. If that makes sense? Hoping I've misunderstood something and can be steered in the right direction :)

fredemmott:
This should be solvable with abstract constants:

```Hack
<?hh // strict

class Foo {
  public function __construct(private string $a, private string $b) {
  }
}

<<__ConsistentConstruct>>
abstract class Bar extends Foo {
  abstract const string ARG_B;

  public function __construct(string $a) {
    parent::__construct($a, static::ARG_B);
  }
}

final class Baz extends Bar {
  const string ARG_B = 'constant b';
}


/* HH_IGNORE_ERROR[1002] pseudomain in strict */
var_dump(new Baz('parameter a'));
```

```
Mac-Pro:hhtest fred$ hh_client
No errors!
Mac-Pro:hhtest fred$ hhvm test.php
object(Baz)#1 (2) {
  ["a":"Foo":private]=>
  string(11) "parameter a"
  ["b":"Foo":private]=>
  string(10) "constant b"
}
```

fredemmott:
worst case, a similar approach works with abstract static methods

bradleyhd:
Ok, updated with the added (arguably-poorly-named) `NoTextEditableToken` and `TextEditableToken` `<<__ConsistentConstruct>>` classes. Thanks for the reviews/ & help thus far!

fredemmott:
Thanks; merged as-is - I'll rename the intermediate classes in a follow-up change

Issue #3, opened 2017-09-22
Add auto-fixing 'no whitespace at end of line' linter


fredemmott:
If there is whitespace at the end of a line, it should be removed.

There's two ways to do this:

 - as an AST linter, looking at each node's first token's leading whitespace, and last token's trailing whitespace
 - making a new 'line based linter' which doesn't use the AST at all, but does use the existing linter base classes. This could also be useful for #2

Line-based is 'good enough' here, though if you use the AST, you could improve on this by excluding HEREDOCs. Please don't handle HEREDOCs or other edge cases if you do a line-based linter.

Either way, please make sure the CLI output is clear when running the linter, both in your terminal, and non-interactively (e.g. when redirecting to a file with `> /tmp/somefile`)

fredemmott:
This was done.

Issue #2, opened 2017-09-22
Add auto-fixing CRLF linter


fredemmott:
If windows newlines are present, the linter should suggest that they're replaced with unix newlines.

There's two ways to do this:
 - as an AST linter, looking at each node's first token's leading whitespace, and last token's trailing whitespace
 - making a new 'line based linter' which doesn't use the AST at all, but does use the existing linter base classes. This could also be useful for #3

Either way, please make sure the CLI output is clear when running the linter, both in your terminal, and non-interactively (e.g. when redirecting to a file with `> /tmp/somefile`)

lexidor:
@fredemmott can this linter blindly assume that `\r\n` at the end of a line is a mistake?
The `NoWhiteSpaceAtTheendOfLineLinter` does this too, even when inside multiline string literals.
If yes, I'll give this a shot.

fredemmott:
Yep, good point - does NoWhiteSpaceAtTheEndOfLineLinter make this irrelevant?

lexidor:
:thinking: Nope, the whitespace at end of line linter doesn't trip on a `\r\n`, but the newline at end of file linter does.
So people using IDE's (like vscode) that have a `LF <-> CRLF` button can't use a windows line ending, without getting a lint error.
Only those of us who use more raw editors (which allow for multiple line endings in one file) could sneak some `\r\n`s in.

The prio is not real high, but when I do get around to it, would we want to extend the capabilities of the whitespace at end of line linter instead of making a new one?

fredemmott:
This should probably be separate - if we ever support windows, we may want 'No whitespace at end of line, but \r\n is end of line"

PR #22, opened 2017-11-03
add NoWhitespaceAtEndOfLineLinter

bradleyhd:
Here's a trailing whitespace AST linter that hopefully addresses #3. Let me know if it needs more tests!

As an aside (I'm pretty new to anything open source), should I be putting the FB copyright headers in new files? Feels weird to be doing that as an outside entity but I'm guessing that's what ultimately needs to happen.


fredemmott:
Thanks - I'm waiting for a definitive answer on the headers.

This looks fine, though - and this isn't a bad thing - is there a reason this is AST-based rather than line-based?

bradleyhd:
Not really! I wrote a line-based one but got to the autofixing part and ended up with something that wrote the file back to disk repeatedly for each error, which felt wrong. Not sure why I didn't just coalesce them in `fixLintErrors` - happy to get it up and running and send it in as a PR. 

I've also got a mixed tabs/whitespace indent line-based linter if there's interest in having that upstream.

Last but not least, do you think #2 (CLRF) should be in the same linter or a separate one?

fredemmott:
It should be a separate linter - it doesn't seem implausible that people would want CRLF removal, but might not want this linter - for example, if it turns out that this has weird interactions with some node types and creates a load of problems in some code bases.

fredemmott:
The standard license header is definitely required.

bradleyhd:
Figured as much :) updated both PRs

fredemmott:
Closing in favor of #23

PR #23, opened 2017-11-07
add line-based trailing whitespace linter

bradleyhd:
As a follow up to #22, here's an attempt at a line-based whitespace linter!


bradleyhd:
whoops, missed the `<<__Override>>`s, updated.

fredemmott:
@bradleyhd: would be a problem if I left thorough reviews of this and #22 for a week or two? I can get to them sooner if needed - I'm asking as there's a rather large pile of stuff on other projects that I need to finish before HHVM 3.23 is released.

bradleyhd:
not a problem at all!

fredemmott:
@bradleyhd : any comments on how these compare when trying them on real code?

bradleyhd:
Sadly not really. I've just been running this against our prod code on an ad-hoc basis, but our current linter already takes care of trailing whitespace so there's never any errors. That said, I like the line-based approach mainly because we have other linters that work well with a line-based model (like detecting inappropriate/insensitive words used anywhere on any line)

src/Linters/AutoFixingLineLinter.php
```diff
@@ -0,0 +1,37 @@
...
+use type Facebook\HHAST\EditableNode;
+use namespace Facebook\HHAST;
+
+abstract class AutoFixingLineLinter<Terr as FixableLineLintError>
+  extends LineLinter implements AutoFixingLinter<Terr> {
+
+  abstract public function getFixedLine(string $line): string;
+
+  final public function fixLintErrors(Traversable<Terr> $errors): void {
+
```
fredemmott: delete empty lines at start of blocks throughout

src/Linters/AutoFixingLineLinter.php
```diff
@@ -0,0 +1,37 @@
...
+    foreach ($errors as $err) {
+
+      $i = $err->getPosition()[0] - 1;
+      invariant($i >= 0, 'line number should never be negative');
+      $original = $lines[$i];
+      $lines[$i] = $this->getFixedLine($original);
+    }
+
+    file_put_contents($this->getFile(), implode("\n", $lines));
```
fredemmott: `\file_put_contents`, and use `Str\join()` - implode is used because of Str\join() being unstable (and changed) before v1.0

src/Linters/LineLinter.php
```diff
@@ -0,0 +1,30 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional
+ * grant of patent rights can be found in the PATENTS file in the same
+ * directory.
+ *
+ */
+
+namespace Facebook\HHAST\Linters;
+
+use Facebook\HHAST\Linters\{BaseLinter};
```
fredemmott: `use type`

src/Linters/NoWhitespaceAtEndOfLineLinter.php
```diff
@@ -0,0 +1,54 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional
+ * grant of patent rights can be found in the PATENTS file in the same
+ * directory.
+ *
+ */
+
+namespace Facebook\HHAST\Linters;
+
+use HH\Lib\Str;
```
fredemmott: `use namespace`

bradleyhd:
Updated, apologies for the lint 🗑

PR #27, opened 2017-11-28
Disabling specific linter errors

johanoskarsson:
Sometimes when linters are enabled there exists lines of code that break the rules. In order to enable the stricter linters while allowing the old code to remain until it can be removed it is useful to be able to disable specific linter errors.

For example let's say we have a linter that bans the usage of certain functions:

`banned_function_call();` would generate a linter error.

However until we can remove it we would be able to disable it like so:

`banned_function_call(); # DisableNameOfLinter`

This PR changes the BaseLinter to add a function that can check if a certain error is disabled or not. It requires that the linter error returns a correct `getPosition()` to a line that has the correct disable string, otherwise it assumes the linter is not disabled.


fredemmott:
What would you think of an inline approach, consistent wit `HH_FIXME[]` and `HH_IGNORE_ERROR[]`? For AST-based linters, this should be in the leading trivia for the node you already have; for line-based linters (eg #23), probably the preceding block, or possibly a per-file/per-block for those - eg

```Hack
<?hh

// HHAST_IGNORE_ALL[WhitespaceAtEndOfLine]

function foo(Traversable<Awaitable<string>> $in): string {
  $out = '';   
  foreach($in as $aw) {
    /* HHAST_FIXME[DontAwaitInALoop] */
    $out .= await $aw;
   }
   return $out;
}
```

Sorry if you've covered this in your implementation - I'm not able to look at that until you sign the CLA.

tests/DisableLineLinterTest.php
```diff
@@ -0,0 +1,36 @@
...
+final class DisableLineLinterTest extends TestCase {
+  use LinterTestTrait;
+
+  protected function getLinter(
+    string $file,
+  ): Linters\BaseLinter {
+    return new Linters\NoWhitespaceAtEndOfLineLinter($file);
+  }
+
+  public function getCleanExamples(): array<array<string>> {
```
fredemmott: this is only array<array<string>> as single-element tuples aren't allowed - you should be concatenating the lines instead of comma-separating them. At the moment I think this is just checking that `<?hh` is clean (and same for the other test)

src/Linters/BaseLinter.php
```diff
@@ -28,4 +28,26 @@ public function __construct(
...
+   */
+  protected function markerIgnoreAll(): string {
+    return 'HHAST_IGNORE_ALL['.$this->getLinterName().']';
+  }
+
+  /**
+   * A user can choose to ignore a specific error reported by this linter using this string as a marker
+   */
+  protected function markerFixMe(): string {
+    return 'HHAST_FIXME['.$this->getLinterName().']';
```
fredemmott: it would be good to support both HHAST_FIXME and HHAST_IGNORE_ERROR - HH_FIXME should be used where it's expected that it will actually be fixed at some point, HH_IGNORE_ERROR when not.

src/Linters/BaseLinter.php
```diff
@@ -28,4 +28,26 @@ public function __construct(
   final public function getFile(): string {
     return $this->file;
   }
+
+  // A simple name for the linter, based on the class name
+  <<__Memoize>>
+  public function getLinterName(): string {
+    $r = new \ReflectionClass($this);
+    return $r->getShortName();
```
fredemmott: How about:

```Hack
return self::class
  |> Str\split($$, '\\')
  |> C\lastx($$)
  |> Str\strip_suffix($$, 'Linter');
```

?

Should both be more performant (no reflection), and lead to a shorter ignore string

src/Linters/BaseASTLinter.php
```diff
@@ -76,13 +77,43 @@ public function getPrettyTextForNode(
...
     }
   }
 
+  /**
+   * Allow users to disable either all linter errors of this type in a file or
+   * specific cases where a linter is used.
+   **/
+  protected function isLinterDisabled(EditableNode $node, LintError $error): bool {
+    // This class of linter errors might be disabled for the entire file, let's find out.
+    if($this->isLinterDisabledForFile($error->getFile())){
```
fredemmott: It would be better to make it so that getLintErrors() calls this and exits before you check any nodes, rather than checking it on every one

johanoskarsson:
Made the change to closer match the syntax you suggested.
The one part I'm not quite satisfied with is the change to NoWhitespaceAtEndOfLineLinter.php to do the individual line check. But since we don't have any other LineLinters in here right now I left it in there, might be possible to push that down further.

src/Linters/BaseASTLinter.php
fredemmott: it looks like you're doing this for some kinds of linters but not others - could you make this happen consistently?

src/Linters/NoWhitespaceAtEndOfLineLinter.php
```diff
@@ -21,12 +21,19 @@
...
+
     foreach ($lines as $ln => $line) {
       for ($i = strlen($line) - 1; $i >= 0; $i--) {
         $char = $line[$i];
         if ($char !== ' ') {
           break;
         }
+
+        // Looks like we have an error. Let's see if we should ignore this one
+        if ($ln-1 >= 0 && Str\contains($lines[$ln-1], $this->markerFixMe())) {
```
fredemmott: AutofixingLineLinter ?  though that might need a refactor that'd be better done separately to make it so that the base does the iteration, and subclasses only look at one line at a time. cc @bradleyhd

bradleyhd: yeah - should probably even be up one more in LineLinter. Alternatively, what do you think about making whether or not the error should be ignored a (derived?) property of the `LintError` itself, and then just `Vec\filter`ing out as necessary in `LinterCLI` ex post facto?

fredemmott: maybe base classes in BaseLinter, but if you’re integrating these into something else without the CLI, you shouldn’t have to explicitly check for ignored ones.

src/Linters/BaseASTLinter.php
```diff
@@ -76,13 +77,28 @@ public function getPrettyTextForNode(
...
     }
   }
 
+  /**
+   * Allow users to disable either all linter errors of this type in a file or
+   * specific cases where a linter is used.
+   **/
+  protected function isLinterDisabled(EditableNode $node, LintError $error): bool {
+    // Is this specific instance of the linter error disabled?
+    $token = $node->getFirstToken();
```
bradleyhd: I'm not sure, but I think there's a scenario where this won't work: for example, if you want to ban function invocations to a particular function and make a linter of type `FunctionCallExpression`, then you won't have the preceding line here if the function call is in a program like
```
// HHAST_FIXME[NoCallsToBanMeLinter]
if (ban_me()) {
  return;
}
```
Not sure if it'd be better to walk `$parents` and look for the right comments or to use `$error->getPosition()` and look at surrounding lines. Or maybe something else?

tests/DisableLineLinterTest.php
bradleyhd: curious if this is true - it looks like there's a similar comma separated approach in `DontAwaitInALoopLinterTest`

src/Linters/BaseASTLinter.php
bradleyhd: `if (/* HHAST_FIXME[NoCallsToBanMeLinter] */ ban_me()) { return; }` attaches the comment to the trail of `(` and not the leading of `ban_me()` 🤔

src/Linters/NoWhitespaceAtEndOfLineLinter.php
bradleyhd: Fair!

src/Linters/BaseASTLinter.php
fredemmott: I think we need to check:
 - own leading
 - leading of every node up to the first statement (or possibly expression)
 - trailing of the previous sibling

tests/DisableLineLinterTest.php
fredemmott: Each test case is separated by a comma; the 'lines' within them are separated with `.`

bradleyhd: wow I really need to up my font size 😬 👓

src/Linters/BaseASTLinter.php
johanoskarsson: @fredemmott could you help me understand this part a bit better? "leading of every node up to the first statement (or possibly expression)". I assume this means we walk up the parents until we hit a certain node? The actual IfStatement, ForStatement etc classes or something else?

fredemmott: e.g. you probably shouldn't be able to use this to suppress every instance of an error in a function

johanoskarsson:
I've made the following changes:
* Reformatted using the command mentioned above
* Updated linter name fetching to not use reflection
* Support for both fixme and ignore error
* Fleshed out the AST disabling check a bit more. I also split this out into a new class as it was getting kind of big
* Added some more tests

However for the final step of pushing down the disabling check int other LineLinter base class I ran into some generics/typing issues. I've been beating my head against these for a while now so figured I would at least push this branch as is and take it from there. I split out the actual refactoring here, but was still unable to solve it: https://github.com/johanoskarsson/hhast/tree/johan_linelinter_refactor
`hhvm vendor/phpunit/phpunit/phpunit tests/NoWhitespaceAtEndOfLineLinterTest.php` would illustrate the problem (some kind of type mismatch makes it not think the getLintErrors is implemented).

fredemmott:
Sorry - I checked out your branch, and both hh_client and the test you mentioned are completely happy - can you give more details about the problem, including what specific version of HHVM you're using?

johanoskarsson:
My bad, I didn't clarify which branch is having the issue. The branch underlying this PR (johan_disabling_linters) should work and pass all tests. I have addressed most of your feedback, however it does not include the change to push down the is disabled check into LineLinter. If you are ok with punting on that change till later we should be good to go here.

For the issue wrt pushing down that code it lives in the other branch (johan_linelinter_refactor). I have put up a gist here that illustrates the problem. I tried with both 3.22.0-1brew (rel) and the latest bottle.
https://gist.github.com/johanoskarsson/07e9cebbe6020c2a67fbb021b5780036
The code is over here: https://github.com/johanoskarsson/hhast/tree/johan_linelinter_refactor

fredemmott:
@johanoskarsson : I can't reproduce that with the exact same build of HHVM, and same commit hash :/

Can you:
 - make sure you don't have any local modifications - you're on a clean checkout of d57ba2cef80c630af50cdbd13cc89cc8ef38c850
 - deleting `~/.hhvm.hhbc` and trying again

src/Linters/BaseLinter.php
```diff
@@ -28,4 +31,46 @@ public function __construct(
...
+  public function markerFixMe(): string {
+    return 'HHAST_FIXME['.$this->getLinterName().']';
+  }
+
+  /**
+   * Is this linter error disabled for the entire file?
+   * Memoized since this should not change per run.
+   */
+  <<__Memoize>>
+  public function isLinterDisabledForFile(string $file_name): bool {
```
fredemmott: Why does this take a file name given `$this->file` is available?

src/Linters/BaseASTLinter.php
```diff
@@ -18,6 +18,7 @@
 
 abstract class BaseASTLinter<T as HHAST\EditableNode, +Terror as ASTLintError<T>> extends BaseLinter {
   private HHAST\EditableNode $ast;
+  private DisableASTLinter $disable_check;
```
fredemmott: camelCase members

src/Linters/DisableASTLinter.php
```diff
@@ -0,0 +1,129 @@
...
+use type Facebook\HHAST\ReturnStatement;
+use type Facebook\HHAST\ThrowStatement;
+use type Facebook\HHAST\TryStatement;
+use type Facebook\HHAST\UnsetStatement;
+use namespace Facebook\HHAST;
+use namespace HH\Lib\Str;
+use namespace HH\Lib\Vec;
+use namespace HH\Lib\C;
+
+class DisableASTLinter {
```
fredemmott: This looks like it could/should be a collection of functions in a namespace - no properties, I don't see a need for this to be instantiable

src/Linters/DisableASTLinter.php
```diff
@@ -0,0 +1,129 @@
...
+use type Facebook\HHAST\EchoStatement;
+use type Facebook\HHAST\GotoStatement;
+use type Facebook\HHAST\ReturnStatement;
+use type Facebook\HHAST\ThrowStatement;
+use type Facebook\HHAST\TryStatement;
+use type Facebook\HHAST\UnsetStatement;
+use namespace Facebook\HHAST;
+use namespace HH\Lib\Str;
+use namespace HH\Lib\Vec;
+use namespace HH\Lib\C;
```
fredemmott: group use would make this clearer, e.g.:

```
use namespace HH\Lib\{C, Str, Vec}
```

Also, alphabetize

src/Linters/DisableASTLinter.php
```diff
@@ -0,0 +1,129 @@
...
+      if (
+        $parent instanceof IControlFlowStatement ||
+        $parent instanceof BreakStatement ||
+        $parent instanceof ContinueStatement ||
+        $parent instanceof EchoStatement ||
+        $parent instanceof GotoStatement ||
+        $parent instanceof ReturnStatement ||
+        $parent instanceof ThrowStatement ||
+        $parent instanceof TryStatement ||
+        $parent instanceof UnsetStatement
```
fredemmott: Doesn't need to be in this PR, but it would be good to change these to implement a common interface - see `ILoopStatement`, `IControlFlowStatement` for an example

tests/DisableASTLinterCaseTest.php
```diff
@@ -0,0 +1,31 @@
...
+final class DisableASTLinterCaseTest extends TestCase {
+  use LinterTestTrait;
+
+  protected function getLinter(string $file): Linters\BaseLinter {
+    return new Linters\CamelCasedMethodsUnderscoredFunctionsLinter($file);
+  }
+
+  public function getCleanExamples(): array<array<string>> {
+    return [
+      ['<?hh /* HHAST_FIXME[CamelCasedMethodsUnderscoredFunctions] */ function CamelCasing() { return null; }'],
```
fredemmott: keep to 80 cols, concatenate multiple strings as in other tests

src/Linters/DisableASTLinter.php
fredemmott: Also, maybe rename to indicate that it suppresses lint errors, rather than implying that it can only disable the entire linter.

src/Linters/NoWhitespaceAtEndOfLineLinter.php
```diff
@@ -21,23 +23,54 @@
...
+    return $errs;
+  }
+
+  public function getLintErrorsForLine(
+    string $line,
+    int $line_number,
+  ): Traversable<FixableLineLintError> {
+    $errs = vec[];
+
+    for ($i = strlen($line) - 1; $i >= 0; $i--) {
```
fredemmott: `Str\length()`

fredemmott:
Okay, so this just needs the previous comments and a rebase now I think :)

johanoskarsson:
This update hopefully contains all the changes requested: cut a few cols shorter, slimmed down the namespaces, removed the disable class and replaced with suppress* versions.

fredemmott:
Thanks so much!

PR #33, opened 2017-12-08
Linelinter refactor for #27

johanoskarsson:
After cloning this fresh all type issues went away...

This is the refactor to push down the looping over the lines of code into LineLinter. This way we have one central place to check for disabling / suppressing linter errors in #27.


src/Linters/AutoFixingLineLinter.php
```diff
@@ -31,4 +31,4 @@
     }
     \file_put_contents($this->getFile(), Str\join($lines, "\n"));
   }
-}
```
fredemmott: is your editor doing this, or hackfmt? are you on windows?

johanoskarsson: This was the hackfmt command from the other PR. I can manually revert those lines.

fredemmott:
We've got an internal issue open for the hackfmt --diff newline at EOF behavior now, and I've filed #35 to mitigate it a bit

PR #51, opened 2018-06-06
Warn if path contains a hhast-lint.json

wlin57:
Hopefully this logic is correct:

- Running 'bin/hhast-lint' works as before
- If the user specifies a directory, check if the directory directly contains a hhast-lint.json. If so, print warning message, but still lint everything in the specified directory.
- If the user specifies a file, just lint the file.
- If the user specifies multiple PATH arguments, print warning message for each case it applies

### DIFF
--- a/src/__Private/LinterCLI.php
+++ b/src/__Private/LinterCLI.php
@@ -162,10 +162,11 @@ final class LinterCLI extends CLIWithArguments {
           $config_file = $path.'/hhast-lint.json';
           if (\file_exists($config_file)) {
             \fwrite(
-              \STDERR,
-              "Warning: PATH arguments contain a hhast-lint.json\n",
+              \STDOUT,
+              "Warning: PATH arguments contain a hhast-lint.json, ".
+              "which modifies the linters used and customizes behavior. ".
+              "Consider 'cd ".$root."; vendor/bin/hhast-lint'\n\n",
             );
-            return 1;
           }
         }
       }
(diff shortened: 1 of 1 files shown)


src/__Private/LinterCLI.php
```diff
@@ -156,6 +156,19 @@ private function lintPath(
...
       }
     } else {
+      foreach ($roots as $root) {
+        $path = \realpath($root);
+        if (\is_dir($path)) {
+          $config_file = $path.'/hhast-lint.json';
+          if (\file_exists($config_file)) {
+            \fwrite(
+              \STDERR,
+              "Warning: PATH arguments contain a hhast-lint.json\n",
```
fredemmott: Logic looks correct, but it would be good to make the message explain why this is a potential problem.

src/__Private/LinterCLI.php
```diff
@@ -156,6 +156,19 @@ private function lintPath(
...
+            );
+            return 1;
```
fredemmott: As this is a warning, not an error, it probably shouldn't exit non-zero, and should probably still run the lints

src/__Private/LinterCLI.php
fredemmott: and what they might want instead (`cd $dir; vendor/bin/hhast-lint`)

src/__Private/LinterCLI.php
wlin57: Should this go to STDOUT then?

PR #54, opened 2018-06-07
Add linter banning PHP equality checks

fredemmott:


fredemmott:
TIL that exists. Let's improve that...

Issue #57, opened 2018-06-07
Add support for JSON-formatted output from hhast-lint


PranayAgarwal:
`hhast-lint` currently produces user readable output for lint errors. It would be helpful to support a `--json` option (similar `hh_client`) to make this parsable by IDEs and other automated tooling.

PR #58, opened 2018-06-08
Add '--json' flag to hhast-lint CLI to write output in JSON format

PranayAgarwal:
Fixes https://github.com/hhvm/hhast/issues/57. This will be very helpful for IDEs, build scripts etc. that need to parse lint results programatically. 

The new `--json` flag skips the pretty printing & automated fixing flow that the linter CLI currently shows, and instead collects all errors in a LSP-compatible shape and outputs it as JSON.

While the new flag works with existing `--verbose` and `--perf` options, the output from those won't be JSON formatted.

**Next steps**

- Update linters to return end position for errors
- Support fixing a single error through CLI args to hhast-lint (may be tricky because the error object will need to be recreated in memory)
- Enable lint CLI to run in LSP mode


src/__Private/LinterCLI.php
```diff
@@ -159,14 +168,42 @@ private function lintPath(
...
-        $had_errors = true;
-      }
+      $all_errors = Vec\concat(
+          $all_errors,
+          $this->lintPath($root_config, $root),
+        );
     }
-    if (!$had_errors) {
+
+    $had_errors = \count($all_errors) > 0;
```
fredemmott: C\count

src/__Private/LinterCLI.php
```diff
@@ -159,14 +168,42 @@ private function lintPath(
...
+                  "character" => $position !== null ? $position[1] : null,
+                ),
+                "end" => shape(
+                  "line" => null,
+                  "character" => null,
+                ),
+              ),
+              "message" => $error->getDescription(),
+              "linter" => \get_class($error->getLinter()),
+              "file" => $error->getFile(),
```
fredemmott: define a shape for this separately

PranayAgarwal:
Yeah this felt weird to me too, but the problem is that existing CLI I/O happens per-linter as they run, while JSON output has to happen at the end after all the results are collected. Any ideas on how to improve this?

fredemmott:
One way would be to move all input/output (including the autofix stuff) out of there; keep it as a generator.

Make separate functions for JSON and CLI mode that take the Traversable and output it

src/__Private/LinterCLI.php
```diff
@@ -38,14 +39,20 @@ protected function getSupportedOptions(): vec<CLIOptions\CLIOption> {
...
+      ),
       $this->getVerbosityOption(),
     ];
   }
 
   private function lintFile(
     LinterCLIConfig $config,
     string $path,
-  ): Traversable<Linters\LintError> {
+    LinterCLIErrorHandler $errorHandler,
```
fredemmott: `$snake_case` for variables (except for properties)

src/__Private/LinterCLI.php
```diff
@@ -71,19 +78,16 @@ private function lintFile(
       }
 
       using (new ScopedPerfCounter($class.'#processErrors')) {
-        $all_errors = Vec\concat(
-          $all_errors,
-          $this->processErrors($linter, $config, $errors),
-        );
+        $errorHandler->processErrors($linter, $config, $errors);
```
fredemmott: why this approach, rather than keeping as a generator and allowing the traverser to deal with them? Either's fine, just want to understand the decision

src/__Private/LinterCLIErrorHandler.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+interface LinterCLIErrorHandler {
+
+  /**
+   * Process a set of errors returned by running an individual linter on a
+   * single file
+   *
+   * @param $linter The linter instance
+   * @param $config Linter configuration for this file's scope
+   * @param $errors Traversable list of errors
```
fredemmott: these param doc cocmmentsdon't add any information

src/__Private/LinterCLIErrorHandlerPlain.php
```diff
@@ -0,0 +1,216 @@
...
+use namespace Facebook\HHAST\Linters;
+use namespace HH\Lib\{C, Str, Vec};
+
+final class LinterCLIErrorHandlerPlain implements LinterCLIErrorHandler {
+
+  private bool $had_errors = false;
+
+  public function __construct(
+    private bool $supportsColors,
+    private bool $isInteractive,
```
fredemmott: we generally consider bool arguments to be an antipattern as it's unclear at the callsite, especially when there's multiple bool params - perhaps a config shape, or enum values instead?

src/__Private/LinterCLIErrorHandlerPlain.php
```diff
@@ -0,0 +1,216 @@
...
+    Linters\BaseLinter $linter,
+    LinterCLIConfig::TFileConfig $config,
+    Traversable<Linters\LintError> $errors,
+  ): void {
+    $class = \get_class($linter);
+    $to_fix = vec[];
+    $colors = $this->supportsColors;
+
+    foreach ($errors as $error) {
+
```
fredemmott: remove

src/__Private/LinterCLIErrorHandlerPlain.php
```diff
@@ -0,0 +1,216 @@
...
+    Linters\FixableLintError $error,
+  ): bool {
+    list($old, $new) = $error->getReadableFix();
+    if ($old === $new) {
+      $this->renderLintBlame($error);
+      return false;
+    }
+
+    $prefix_lines = ($code, $prefix) ==>
+      \explode("\n", $code)
```
fredemmott: Str\split

src/__Private/LinterCLIErrorHandlerPlain.php
```diff
@@ -0,0 +1,216 @@
...
+      |> Vec\map(
+        $$,
+        $line ==> $prefix.$line,
+      )
+      |> \implode("\n", $$);
```
fredemmott: Str\join

src/__Private/LinterCLIErrorHandlerPlain.php
```diff
@@ -0,0 +1,216 @@
...
+    \printf(
+      "  Code:\n%s%s%s\n",
+      $colors ? "\e[33m" : '',
+      \explode("\n", $blame)
+      |> Vec\map(
+        $$,
+        $line ==> '  >'.$line,
+      )
+      |> \implode("\n", $$),
+      $colors ? "\e[0m" : '',
```
fredemmott: Str split joi nagain

src/__Private/LinterCLI.php
PranayAgarwal: It's because linters can fix all errors from a file in the same call, so passing a single traversable list of errors won't work. I found this approach cleaner than generating `Traversable<Traversable<Error>>` and using that.

src/__Private/LinterCLI.php
```diff
@@ -77,19 +84,16 @@ private function lintFile(
...
       }
     }
-    return $all_errors;
   }
 
   private function lintDirectory(
     LinterCLIConfig $config,
     string $path,
-  ): Traversable<Linters\LintError> {
+    LinterCLIErrorHandler $errorHandler,
```
fredemmott: $error_handler for locals/parameters

src/__Private/LinterCLIErrorHandler.php
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Check whether this handler has seen any (unfixed) lint errors
+   */
+  public function hadErrors(): bool;
+
+  /**
+   * Print the final output of the linter after all errors have been processed
+   */
+  public function print(): void;
+}
```
fredemmott: newline at EOF

PranayAgarwal:
There should be linters for all of these :)

fredemmott:
Please rebase future changes instead of merges - makes it much easier to manage :)

PR #59, opened 2018-06-08
Add option to disable all linters for a pattern with an override

wlin57:


src/__Private/LinterCLIConfig.php
```diff
@@ -146,27 +156,33 @@ public function getConfigForFile(string $file_path): self::TFileConfig {
...
-        $autofix_blacklist,
-        $override['disabledAutoFixes'] ?? vec[],
-      );
-      $no_autofixes = $no_autofixes ||
-        ($override['disableAllAutoFixes'] ?? false);
+      if ($override['disableAllLinters'] ?? false) {
+        $linters = $normalize(vec[]);
+        $autofix_blacklist = $normalize(vec[]);
+        $linters = $assert_types($linters);
+        $autofix_blacklist = $assert_types($autofix_blacklist);
```
fredemmott: The noramlize/assert_types lambdas seems unnecessary  - this is equivalent to:

```
return shape(
  'linters' => keyset[],
  'autoFixBlacklist' => keyset[],
);
```

also, setting the autofixblacklist should be unneccessary if we're not running any linters

src/__Private/LinterCLIConfig.php
```diff
@@ -146,27 +156,33 @@ public function getConfigForFile(string $file_path): self::TFileConfig {
...
-        $override['disabledLinters'] ?? vec[],
-      );
-      $autofix_blacklist = Vec\concat(
-        $autofix_blacklist,
-        $override['disabledAutoFixes'] ?? vec[],
-      );
-      $no_autofixes = $no_autofixes ||
-        ($override['disableAllAutoFixes'] ?? false);
+      if ($override['disableAllLinters'] ?? false) {
+        $linters = $normalize(vec[]);
```
kmeht: Why not empty keyset?

src/__Private/LinterCLIConfig.php
```diff
@@ -146,27 +156,33 @@ public function getConfigForFile(string $file_path): self::TFileConfig {
...
+        $autofix_blacklist = $normalize(vec[]);
```
kmeht: ditto

src/__Private/LinterCLIConfig.php
```diff
@@ -146,27 +156,33 @@ public function getConfigForFile(string $file_path): self::TFileConfig {
...
+        $linters = $assert_types($linters);
+        $autofix_blacklist = $assert_types($autofix_blacklist);
+        return shape(
+          'linters' => $linters,
+          'autoFixBlacklist' => $autofix_blacklist,
+        );
+      } else {
```
kmeht: This doesn't need to be in an `else` block, since you return early in the case of disabling all linters.

src/__Private/LinterCLIConfig.php
wlin57: ah, i was looking for this.

src/__Private/LinterCLIConfig.php
```diff
@@ -160,13 +165,12 @@ public function getConfigForFile(string $file_path): self::TFileConfig {
       );
       $no_autofixes = $no_autofixes ||
         ($override['disableAllAutoFixes'] ?? false);
-    }
+      }
```
kmeht: Revert before landing.

src/__Private/LinterCLIConfig.php
```diff
@@ -160,13 +165,12 @@ public function getConfigForFile(string $file_path): self::TFileConfig {
...
 
     $normalize = $list ==> Keyset\map(
       $list,
       $linter ==> $this->getFullyQualifiedLinterName($linter),
     );
-
```
kmeht: Ditto.

PR #60, opened 2018-06-10
Improve handling of multiple short options by supporting short syntax

### LINKED ISSUE #1: CLI: improve handling of multiple short options
https://github.com/hhvm/hhast/blob/master/src/__Private/CLIBase.php

It's normal to be able to combine multiple short options into a single one - for example, `-a -b` should be equivalent to `-ab`.

Complications:
 - this shouldn't be true for short options which require an argument - for example, if b needs an option, `-a -b foo` is required  - not `-abc foo` or `-ab foo`
 - this will require banning multi-character short options - you can do this in the CLIOption base class with an invariant
 - that means that `-vv` will need to be removed from `bin/hhast-lint`; this can be done by changing `-v` to increment `$verbosity`, instead of setting it to 1

YuryBandarchuk16:
Fix #1

### DIFF
--- a/src/__Private/CLIBase.php
+++ b/src/__Private/CLIBase.php
@@ -217,10 +217,7 @@ abstract class CLIBase {
         break;
 
       case CLIOptions\CLIOptionType::SHORT:
-        $arg_length = Str\length($arg);
-        for ($index = 0; $index < $arg_length; $index++) {
-          $options[] = $arg[$index];
-        }
+        $options = Vec\concat($options, Str\chunk($arg));
         break;
 
       default:
@@ -230,21 +227,17 @@ abstract class CLIBase {
     $is_single_option = C\count($options) === 1;
     if ($is_single_option) {
       $value = null;
-      $option = C\firstx($options);
+      $option = C\onlyx($options);
       if (Str\contains($option, '=')) {
-        $parts = \explode('=', $option);
+        $parts = Str\split($option, '=');
         $opt = $parts[0];
-        $value = \implode('=', Vec\drop($parts, 1));
+        $value = Str\join(Vec\drop($parts, 1), '=');
       }
       $result = dict[
         $option => $value,
       ];
     } else {
-      $result = Dict\pull(
-        $options,
-        $_ ==> null,
-        $option ==> $option
-      );
+      $result = Dict\fill_keys($options, null);
     }
 
     return $result;
--- a/src/__Private/CLIBase.php
+++ b/src/__Private/CLIBase.php
@@ -208,7 +208,7 @@ abstract class CLIBase {
     exit(1);
   }
 
-  final private static function extractOptions(string $arg, CLIOptions\CLIOptionType $type): dict<string, ?string> {
+  private static function extractOptions(string $arg, CLIOptions\CLIOptionType $type): dict<string, ?string> {
     $options = vec[];
 
     switch ($type) {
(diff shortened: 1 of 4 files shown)


src/__Private/CLIBase.php
```diff
@@ -206,6 +208,48 @@ public function __construct(
...
+      if (Str\contains($option, '=')) {
+        $parts = \explode('=', $option);
+        $opt = $parts[0];
+        $value = \implode('=', Vec\drop($parts, 1));
+      }
+      $result = dict[
+        $option => $value,
+      ];
+    } else {
+      $result = Dict\pull(
```
kmeht:
```
$result = Dict\fill_keys($options, null);
```

src/__Private/CLIBase.php
```diff
@@ -206,6 +208,48 @@ public function __construct(
...
+        break;
+
+      default:
+        break;
+    }
+
+    $is_single_option = C\count($options) === 1;
+    if ($is_single_option) {
+      $value = null;
+      $option = C\firstx($options);
```
kmeht: Optional: `C\onlyx`.

src/__Private/CLIBase.php
```diff
@@ -206,6 +208,48 @@ public function __construct(
...
+    switch ($type) {
+      case CLIOptions\CLIOptionType::LONG:
+        $options[] = $arg;
+        break;
+
+      case CLIOptions\CLIOptionType::SHORT:
+        $arg_length = Str\length($arg);
+        for ($index = 0; $index < $arg_length; $index++) {
+          $options[] = $arg[$index];
+        }
```
kmeht:
```
$options = Vec\concat($options, Str\chunk($arg));
```

src/__Private/CLIBase.php
```diff
@@ -206,6 +208,48 @@ public function __construct(
...
+      default:
+        break;
+    }
+
+    $is_single_option = C\count($options) === 1;
+    if ($is_single_option) {
+      $value = null;
+      $option = C\firstx($options);
+      if (Str\contains($option, '=')) {
+        $parts = \explode('=', $option);
```
kmeht: `Str\split($option, '=')`

src/__Private/CLIBase.php
```diff
@@ -206,6 +208,48 @@ public function __construct(
...
+        $opt = $parts[0];
+        $value = \implode('=', Vec\drop($parts, 1));
```
kmeht: `Str\join(Vec\drop($parts, 1), '=')`

src/__Private/CLIBase.php
```diff
@@ -206,6 +208,48 @@ public function __construct(
...
+        break;
+
+      case CLIOptions\CLIOptionType::SHORT:
+        $arg_length = Str\length($arg);
+        for ($index = 0; $index < $arg_length; $index++) {
+          $options[] = $arg[$index];
+        }
+        break;
+
+      default:
```
kmeht: Can remove this, Hack can do exhaustiveness checking on enums.

YuryBandarchuk16: Actually, I omitted one enum value so could not remove default branch

fredemmott:
I'm going to hold off on merging this for a few days; I'm doing some work that would make the CLI library testable, which I think is needed before merging this.

fredemmott:
This is going to be resubmitted against hh-clilib

PR #62, opened 2018-06-11
Remove PerfCounter and ScopedPerfCounter

fredemmott:
- `--xhprof` provides better data
- this improves code readability

Downside: `--xhprof` is significantly slower at runtime. I think it's
worth it to avoid having profiling calls spread throughout everything.


fredemmott:
@bradleyhd , @johanoskarsson : what do you think about this?

johanoskarsson:
I'm fine with just using xhprof for the linting profiling. I'm not sure if this would make it harder to monitor/performance tune CI runs of the linter but we haven't had any issues with these yet.

Issue #25, opened 2017-11-20
Add support for the language service protocol


fredemmott:

fredemmott:
cc @bradleyhd, @PranayAgarwal :p

arxanas:
Here's what would need to be done to add support for the language server protocol.

You can review the protocol at https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md

Note this unfortunate reality: when I last checked, many editors didn't support multiple language services for a single file type. When I looked at the LSP in Neovim it looked like there's no way to get LSP support for both Hack and the linting framework. (There would be a lot of use for some sort of LSP multiplexer that runs multiple language services under one facade for these editors.) Nuclide supports multiple language services. VS Code probably does but I haven't checked.

## Basic LSP
* Implement JSON-RPC as a transport layer. This would just mean that the script is running with stdin and stdout reading and writing a basic JSON protocol. The details of JSON-RPC as relevant to the LSP are in the LSP protocol spec.
* Handle the `initialize` method, which is sent when the editor starts up. Make sure that the server advertises that it cares about open/close notifications and change notifications in `TextDocumentSyncOptions`. (I spent a couple days trying to figure out why I wasn't receiving document changes on the server because I didn't set this option.)
* Handle the `shutdown` method and `exit` notification.

## Synchronization

* The language server will keep the states of all the documents in memory. You need to handle `textDocument/didOpen`, `textDocument/didChange`, and `textDocument/didClose` notifications and mirror whatever changes are being made in the editor.

## Lints

* For lint issues, send the `textDocument/publishDiagnostics` notification from the server to the client whenever you've calculated lint issues to display.
* Make sure to send the empty set of diagnostics to a document that you no longer want to show diagnostics for. (That is, if you only want lint issues to show on your open files, then you'll need to clear out the diagnostics when the user closes the file. Fortunately this would be a simple hook into `textDocument/didClose`.)

## Refactorings

* For refactorings, it's unfortunately a bit more complicated. The LSP has a concept of "code actions". You may have seen the lightbulb icon in VS Code: these are code actions.
* Nuclide will request code actions whenever you hover over a diagnostic. I'm not sure about the rules in VS Code, but I think it's in more cases. In particular, VS Code will probably send a code actions request when you make a selection in the document. You probably don't need to care about that unless you want to implement something like "extract into method", which takes a selection and extracts it into a new method in the class. I haven't tried to use the range yet, so I might be wrong about the details.
[...]

PR #66, opened 2018-06-15
LSP interface definitions, POC in LinterCLI

fredemmott:
This builds on top of #65, and only implements the `initialize` message. It also depends on https://github.com/hhvm/hh-clilib/commit/b8ac1045c4333bca3444a454d7f0297f82c2af9f

You probably want to look at this commit-at-a-time, instead of all the changes.

It includes *all* the LSP spec interfaces, not just the ones we hope to use in HHAST.

refs #25

in vscode:

```
[Info  - 4:02:28 PM] Connection to server got closed. Server will restart.
Waiting for init
Got Init
```

### DIFF
--- a/src/__Private/LintRunLSPErrorHandler.php
+++ b/src/__Private/LintRunLSPErrorHandler.php
@@ -47,7 +47,7 @@ final class LintRunLSPErrorHandler implements LintRunErrorHandler {
           $start = shape('line' => $position[0] - 1, 'character' => $position[1]);
           return shape(
             'range' => shape('start' => $start, 'end' => $start),
-            'severity' => LSP\DiagnosticSeverity::ERROR,
+            'severity' => LSP\DiagnosticSeverity::WARNING,
             'code' => \get_class($linter)
               |> Str\split($$, "\\")
               |> C\lastx($$)
@@ -68,5 +68,8 @@ final class LintRunLSPErrorHandler implements LintRunErrorHandler {
           $encoded,
         ),
       );
+  }
+
+  final public function printFinalOutput(): void {
   }
 }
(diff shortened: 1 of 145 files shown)


fredemmott:
oh hello

screen shot 2018-06-15 at 10 08 57 pm

PranayAgarwal:
I don't think the LSP spec (in JavaScript) differentiates between nullable and optional fields in an object. Since the Hack ones are all marked optional, will there be problems if it receives `field => null` from the server?

PranayAgarwal:
Actually I may be wrong, since TypeScript does allow `name?: bool | null`

fredemmott:
No, especially as I'm using typecoerce instead of typeassert

src/__Private/LintRunErrorHandler.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace Facebook\HHAST\__Private;
+
+use type Facebook\CLILib\OutputInterface;
+use namespace Facebook\HHAST\Linters;
+
+/**
+ * Interface for creating custom handlers for lint errors that are found during
+ * a hhast-lint CLI run. Some examples of hander actions are generating human
```
PranayAgarwal: Missed this in the last PR, but I don't think this file only applies to CLI runs?

src/__Private/LintRunLSPErrorHandler.php
```diff
@@ -0,0 +1,80 @@
+<?hh // strict
```
PranayAgarwal: File name should ideally be `LintRunErrorHandlerLSP` to be consistent with the others

PranayAgarwal: Or really just use some more consistency in general (e.g. `Lint` vs `Linter`)

src/__Private/LintRunErrorHandler.php
fredemmott: yep

PR #70, opened 2018-06-19
Add linter for `use` instead of `use type` or `use namespace`

fredemmott:


src/Linters/UseStatementWIthoutKindLinter.php
```diff
@@ -0,0 +1,128 @@
...
+    INamespaceUseDeclaration $node,
+    vec<EditableNode> $_context,
+  ): ?FixableASTLintError<INamespaceUseDeclaration> {
+    if ($node->hasKind()) {
+      return null;
+    }
+
+    return new FixableASTLintError(
+      $this,
+      "Use `use type` or `use namespace`",
```
kmeht: Isn't `use function` a thing too?

fredemmott: Yep, however `use Foo` is equivalent to `use type Foo; use namespace Foo;` - it doesn't include `use function` or `use const`

fredemmott:
@wlin57 : the test trait applies all suggested fixes, and compares the output: https://github.com/hhvm/hhast/blob/master/tests/AutoFixingLinterTestTrait.php#L43

PR #76, opened 2018-06-20
allow out-of-order responses

fredemmott:
This allows lint results to be published for specific files (e.g.
changed files) before the main lint run has finished.

depends on (and includes) #75


fredemmott:
@jano: can you look over bd16758 please?

fredemmott:
failures are due to parents needing a restack :(

src/__Private/LSPImpl/Server.php
```diff
@@ -83,9 +84,19 @@ protected function getSupportedClientNotifications(
...
+          $body = await $this->readMessageAsync();
+          $poll->add(async {
+            await $this->handleMessageAsync($body);
+          });
+        }
+      },
+    );
+    foreach ($poll await as $_) {
+      // do nothing
     }
```
jano: If all you want to do is to await everything, you don't need AsyncPoll. This is equivalent to:

```
$awaitables = vec[];
while (!$stdin->isEof()) {
  $body = await $this->readMessageAsync();
  // start handler
  $awaitables[] = $this->handleMessageAsync($body);
}

// wait for all handlers to finish
await Vec\from_async($awaitables);
```

fredemmott: With that, nothing from `$awaitables` will execute until $stdin reaches EOF

jano: It should, await does not trigger execution, it just waits for completion.

Unless you are not inside the scheduler yet. If you just HH\Asio\join(mainAsync()), then mainAsync() will run eagerly outside of scheduler until getting blocked. If the current implementation of readMessageAsync() does not suspend execution and runs fully eagerly, then we will not be inside HH\Asio\join() yet by the time we run handleMessageAsync().

The proper way to enter the scheduler is via something like Asio::enterAsyncEntryPoint() (see www). Not sure if we have an OSS equivalent of that.

fredemmott: Sorry, to clarify:

If I use your code, nothing is guaranteed to *complete* until after `$stdin->eof()`. That breaks this functionality.

The purpose of this commit is to allow:

```
> receive request 1 on stdin
> receive request 2 on stdin
< send response 2 on stdout
> receive request 3 on stdin
< send reponse 2 on stdout
< send response 1 on stdout
> receive request 4 on stdin
< send response 4 on stdout
```

... or any other interleaving. For extra context, stdin should never reach EOF while the program is running.

fredemmott: Or are you saying that the scheduler is guaranteed to eventually fully execute all of these awaitables even if we never exit the while loop, as long as readMessageAsync suspends execution?

jano: If the handleMessageAsync() for message 2 completes before reading the one for message 3, it will complete irrespective of whether it was awaited or not. As long as we are inside the scheduler.

```
<?hh

async function work(int $sleep, string $task) {
  $time = microtime(true);
  echo "starting $task @ $time\n";
  await SleepWaitHandle::create($sleep * 1000000);
  $time = microtime(true);
  echo "ending $task @ $time\n";
}

async function run() {
  $input = vec[tuple(0, 10), tuple(1, 1), tuple(2, 3)];
  $awaitables = vec[];
  foreach ($input as $idx => list($read_time, $process_time)) {
    await work($read_time, "read $idx");
    $awaitables[] = work($process_time, "process $idx");
  }

  echo "waiting for everything to finish\n";
  foreach ($awaitables as $awaitable) {
    await $awaitable;
  }
  echo "done\n";
}

HH\Asio\join(async {
  await RescheduleWaitHandle::create(0, 0);
  await run();
});
```

```
starting read 0 @ 1529698425.318
ending read 0 @ 1529698425.318
starting process 0 @ 1529698425.318
starting read 1 @ 1529698425.318
ending read 1 @ 1529698426.3181
starting process 1 @ 1529698426.3182
starting read 2 @ 1529698426.3182
ending process 1 @ 1529698427.3184
ending read 2 @ 1529698428.3183
starting process 2 @ 1529698428.3184
waiting for everything to finish
ending process 2 @ 1529698431.3185
ending process 0 @ 1529698435.3181
done
```

fredemmott: Is that guaranteed?

I was under the impression that everything the scheduler does before an `await` is considered undefined behavior, and shouldn't be depended on

fredemmott: *everything the scheduler does regarding a particular waithandle, until that waithandle is awaited

fredemmott:
Replaced by the much-smaller https://github.com/hhvm/hhast/pull/76/commits/3f6ceab86cb528163c6ae505e9c4c784fa3e3c9d - @jano, can you confirm that's safe?

fredemmott:
that works. I thought we generally didn't like doign that because of backtraces or similar.

PR #93, opened 2018-06-28
A better^Wworse^Wdifferent way to deal with non-UTF8 sources

fredemmott:
JSON can't represent non-UTF8 strings, so this isn't fully fixable.

However, we don't actually need the text, only the metadata, so it's
pointless trying to convert it nicely.

Do something not nice, but that will always work.


fredemmott:
also cc @alexeyt, @arxanas

Issue #91, opened 2018-06-22
Add PHPUnit assertion => FBExpect migration


fredemmott:
Replace `$this->assertEquals()` and similar with `expect($a)->tobePHPEqual($b)`

In addition to being a pre-requisite to replacing PHPUnit, this solves real issues in tests introduced by Hack array migration: `assertEquals()` is always true if both sides are vecs for example, even if the vecs have completely different elements.

This should be a multi-step migration (see https://github.com/hhvm/hhast/blob/master/src/Migrations/ImplicitShapeSubtypesMigration.php for an example). It should:

1. add `use function Facebook\FBExpect\expect` if it isn't already present. This should be (in preference order, depending on the file)
   - next to existing `use function` if present
   - next to other `use` declarations if present
   - immediately after a `namespace` declaration (or just inside the block) if there is one
   - at the top of the file, after any leading comments that aren't docblock comments (copyright notices are a common example)
2. change understood assertions (i.e. only change assertFoo if we know that expect has a replacement for Foo, not just everything that matches that pattern)

This shouldn't change the base class from PHPUnit.

fredemmott:
Here's the basics of just assertEquals - doesn't handle whitespace, comments, generating nice code etc, but gets the semantics right:

```
<?hh // strict
/*
 *  Copyright (c) 2017-present, Facebook, Inc.
 *  All rights reserved.
 *
 *  This source code is licensed under the MIT license found in the
 *  LICENSE file in the root directory of this source tree.
 *
 */

namespace Facebook\HHAST\Linters;

use type Facebook\HHAST\{
  EditableNode,
  EditableList,
  FunctionCallExpression,
  BinaryExpression,
  MemberSelectionExpression,
  LeftParenToken,
  RightParenToken,
  NameToken,
  EqualToken,
  DelimitedComment,
  ListItem,
  CommaToken,
  WhiteSpace,
};
use function Facebook\HHAST\Missing;
use namespace Facebook\TypeAssert;
use namespace HH\Lib\{C, Str, Vec};

final class NoAssertEqualsLinter
  extends AutoFixingASTLinter<FunctionCallExpression> {

  <<__Override>>
  protected static function getTargetType(): classname<FunctionCallExpression> {
    return FunctionCallExpression::class;
  }

  <<__Override>>
  public function getLintErrorForNode(
    FunctionCallExpression $node,
    vec<EditableNode> $parents,
  ): ?FixableASTLintError<FunctionCallExpression> {
[...]

PR #102, opened 2018-07-10
Assert to Expect migration

wlin57:
- Migrates assert calls to their respective expect function (let me know if I missed any others that directly translate)
- Adds 'use function Facebook\FBExpect\expect' according to #91 
- Doesn't handle all of the cases with comments correctly...working on it

Will add examples of each case listed in #91 , or perhaps refactor how MigrationsTest works (currently can only test one .in and .expect file)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,23 +1,23 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
     "content-hash": "b8981db4f6f199cc6229bc4daef10b13",
     "packages": [
         {
             "name": "facebook/hh-clilib",
-            "version": "v1.1.1",
+            "version": "v1.1.2",
             "source": {
                 "type": "git",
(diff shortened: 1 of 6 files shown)


kmeht:
Is this meant to migrate tests internally? If so, the `assert*` functions are static methods, not instance methods.

fredemmott:
Test failure looks like a Travis infra fail, triggered a re-run.

fredemmott:
> Adds 'use function Facebook\FBExpect\expect' according to #91

We also don't want this internally except for open source code like the HSL (we have `expect()` in the root namespace that returns a subclass).

Probably best to just comment out this part when running internally rather than add the complexity to fully support making it optional.

fredemmott:
I thought there was a comment saying this was incomplete - is it ready now?

Also, can you test this by sending a PR against an open source FB project (maybe hhast itself?) that includes the results of this migration please

wlin57:
Almost, it doesn't handle some edge cases with comments. However, I think it will work fine for an open source FB project. I'll try definition-finder since hhast doesn't really have any assert calls in its tests.

wlin57:
The latest commit resolves some bugs (found when testing against definition-finder) and updates tests accordingly. The thing with open source is that there's always (?) a copyright comment or some form of namespace or use statement. Unless there are multiple successive non-docblock comments and no namespace/use statements at all, this migration should work as expected for open source.

PR #103, opened 2018-07-10
Add an atom-ide extension

fredemmott:
Works fine with nuclide.

Went for typescript given that the vscode-hack extension's HHAST support
is pretty much identical, and I'd rather not have to use different
syntaxes for changes to each.


fredemmott:
@pittsw, @ljw1004, @arxanas: I'd appreciate your review; for context, `hhast-lint` is similar to aurora, and also speaks LSP.

Equivalent vscode support is in https://github.com/PranayAgarwal/vscode-hack/blob/master/src/LSPHHASTLint.ts

I've not bothered with a separate killswitch in this as it's a dedicated extension.

There's also a question of what the desired experience is for Hack itself in Atom; some options are:
 - "just use Nuclide"
 - split out "nuclide-hack" so it's installable with just atom-ide, not the rest of Nuclide
 - given how tiny nuclide-hack would be without the legacy stuff and nuclide-specific extensions, maybe it just makes more sense for me to make an `atom-ide-hack` ? I suspect that would be smaller than this extension

pittsw:
Looks good to me.

PR #105, opened 2018-07-11
Add www assert to expect migration

wlin57:
- Migrates assertFoo and self::assertFoo to expect in order for this to work in www
- Two versions: one for OSS and one to be used on www
- Don't need 'use function Facebook\FBExpect\expect' in www, so those steps are removed for www version


src/__Private/MigrationCLI.php
```diff
@@ -48,9 +49,14 @@ class MigrationCLI extends CLIWithRequiredArguments {
...
-        'Change assert calls to expect',
-        '--expect',
+        () ==> { $this->migrations[] = ExternalAssertToExpectMigration::class; },
+        'Change assert calls to expect (external)',
+        '--oss-expect',
+      ),
+      CLIOptions\flag(
+        () ==> { $this->migrations[] = InternalAssertToExpectMigration::class; },
+        'Change assert calls to expect (internal)',
+        '--www-expect',
```
fredemmott: Internal/External/WWW are only informative for FB engineers, which aren't the primary audicence of this tool; renaming to static/member might be best - that said, while looking through the PHPUnit docs tro find out when this changed, it didn't: apparently the convention is to call them 'incorrectly' (see http://phpunit.readthedocs.io/en/7.1/assertions.html)

This makes me think that we should probably support both static and non-static invocations in the same migration (but possible as multiple steps)

wlin57:
Now supports both static and non-static invocations in the same migration (and steps).

PR #106, opened 2018-07-13
Add auto-fixing linter for unused use statements

fredemmott:
- does not support `use constant` as that's too involved
- test run will fail, as I've omitted the fixes to HHAST itself - those are here: https://gist.github.com/fredemmott/1c66c2d3066be80697d98110c4ebca5a

I've not included the fixes in this PR as they make the PR too big to be reviewable - will add them before merging.

### DIFF
--- a/src/Linters/UnusedUseClauseLinter.php
+++ b/src/Linters/UnusedUseClauseLinter.php
@@ -106,9 +106,8 @@ final class UnusedUseClauseLinter
       }
 
       if ($kind instanceof ConstToken) {
-        if (!C\contains($used['constants'], $as)) {
-          $unused[] = tuple($as, $clause);
-        }
+        // unsupported
+        continue;
         continue;
       }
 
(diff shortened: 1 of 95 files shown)


src/Linters/UnusedUseClauseLinter.php
```diff
@@ -0,0 +1,228 @@
...
+
+    return new FixableASTLintError(
+      $this,
+      C\count($unused) === 1
+        ? ('`'.C\firstx($unused)[0].'` is not used')
+        : (
+            $unused
+            |> Vec\map($$, $p ==> '`'.$p[0].'`')
+            |> Str\join($$, ', ')
+            |> $$.' are not used'
```
kmeht: Could probably just switch on the count to swap "are" and "is"

src/Linters/UnusedUseClauseLinter.php
```diff
@@ -0,0 +1,228 @@
...
+        if (!C\contains($used['functions'], $as)) {
+          $unused[] = tuple($as, $clause);
+        }
+        continue;
+      }
+
+      if ($kind instanceof ConstToken) {
+        // unsupported
+        continue;
+        continue;
```
kmeht: remove

Issue #6, opened 2017-09-25
Add `->replaceWithCode(EditableNode, string)` or similar


fredemmott:
It's somewhat more convenient than actually having to construct a replacement AST for simple fixes/migrations. This should use a manually created EditableNode subclass with no children, and user-specifiable text.

The main downside is that it doesn't stack correctly - if a later migration step needs to modify the same code, it will fail. This should be covered in the documentation.

lexidor:
I am not so sure about this part in 2023: _This should use a manually created EditableNode subclass with no children, and user-specifiable text._ This was likely a better call in 2017.

Now that hhast can parse code without needing to `await` anything, there is something smarter we could do.

```
Foreach $outgoing_node get ancestors (including $outgoing_node) as $a:
  If $a is IExpression or IStatement:
    Splice $replacement_string into $a get code
    Parse function x() { /* statement text here */ }
    Extract $replacement_node from output and return
    
  If $a is declaration:
    Prune all runnable code from $a (method bodies etc, 
      since $outgoing_node is not runnable, see previous if.)
    Splice $replacement_string into $a get code
    Parse said code
    Extract $replacement_node from output and return

If none match, slow path: splice into Script get code parse and extract
```

This would mitigate:
```
The main downside is that it doesn't stack correctly - if a later migration step needs to modify the same code, it will fail.
```

Tip, serializing and reparsing a tree might change the structure. Other autofixes might have replaced a node with an "incorrect" node with the right text, making a trivial diff again $a impractical. If that's the case, serialize and reparse the unmodified $a to get back to simple node-by-node diffing.

PR #108, opened 2018-08-03
Add HHAST migration to convert PHP stdlib calls to HSL

ssandler:
Posting this to start a discussion about how to make this ready for wider use. We're using this to migrate functions from PHP standard library calls to HSL. It's got some rough edges, but has been useful enough to run against much of the codebase without major issues.

Things on my mind to improve:

1. Clean up `getUseDeclaration` to automatically group multiple HSL namespaces. Also is generating the code as a string ok here? I originally did it with the AST generating each node at a time... but just generating one line of code required instantiating 49 classes, so readability suffers.
2. Detect the right argument order for `implode()` conversions, since that's ambiguous
3. Add more functions to this, if any are missing
4. Automatically detect whether to use `Str\replace_every()` instead of `Str\replace()` depending on whether the the args are KeyedContainers vs. strings. Currently I've been fixing these cases by hand after running the migrator.
5. Add unit tests
6. Make the bits that detect null vs. false comparisons smarter. Possibly integrate with `hh_client` and look for type errors that can be fixed by changing `false` to `null`. Right now it works nicely if the result of the function call like `Str\search` is compared with `false` immediately, but if it's saved off and used elsewhere it has to be fixed by hand. Luckily, the new hh_client errors about always-false expressions help a ton with that.
7. Detect negative length arguments to `Str\slice` if possible, since those throw exceptions. That's been a sticking point for us which we've had to manually keep an eye out for.
8. Whitespace formatting on this file. I ran `hackfmt` with no other args. Is that the norm? The formatting on `PHP_HSL_REPLACEMENTS` is inconsistent from one element to the next. Perhaps I should bump up the line width so it puts all the configs on one line?
9. Insert your ideas here


ssandler:
Just pushed a whole lot of updates, addressing all the issues I posted above, and a bunch more:

- Added unit tests. I put a bunch of echo commands in these example files, and verified that when actually executing the code the "expected" and "in" file produce the same results (other than `array` vs. `vec`).
- Automatically group namespace use declarations
- Use hh_client to find the types of nodes, and use that to fix the argument order for `implode()` which is ambiguous, as well as `str_replace` with array vs. string inputs
- Rewrite negative length arguments to `Str\slice()` so that they don't throw runtime exceptions. This was harder to do for `Str\splice()` and we don't use the latter much, so I didn't handle negative lengths for splice, just skip those invocations if found.
- Add a whole lot of new functions, covering everything I could find in the HSL reference that has a straightforward equivalent
- `min()` vs. `minva()` based on argument count, and same for max/maxva

I think this is about as much functionality as I want on this initial PR. More ideas for future improvements:
- batch up the requests to `hh_client`, since it can get slow if the few functions we use it for are called a lot in the file.
- handle other built-ins that are harder to convert over, like strncmp
- handle `isset()`, `empty()` in some intelligent way
- tie this code up to an auto-fixing linter
- convert `in_array()`, `array_key_exists()` to use `C\contains()`, `C\contains_key()`

src/find_type_for_node.php
```diff
@@ -0,0 +1,78 @@
...
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST\__Private;
+;
+
+
```
fredemmott: Delete one of these blank lines. https://github.com/hhvm/hhast/issues/109

src/find_type_for_node.php
```diff
@@ -0,0 +1,78 @@
...
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST\__Private;
+;
```
fredemmott: Delete. #110

src/find_type_for_node.php
```diff
@@ -0,0 +1,78 @@
...
+  ),
+  ...
+);
+
+function find_type_for_node(
+  EditableNode $root,
+  EditableNode $node,
+  string $path,
+): ?string {
+
```
fredemmott: delete (#111)

src/find_type_for_node.php
```diff
@@ -0,0 +1,78 @@
...
+    '--json',
+    '--from',
+    'hhast',
+    '--type-at-pos',
+    $line.':'.($offset + 1),
+    '<',
+    \escapeshellarg($path),
+  ];
+  $output = \tempnam(\sys_get_temp_dir(), 'hhast-temp');
+  \exec(Str\join($command, ' ').' >'.\escapeshellarg($output));
```
fredemmott: make this function async, and use/extend `__Private\execute_async` if possible

src/find_type_for_node.php
```diff
@@ -0,0 +1,78 @@
...
+  }
+  ;
+
+  $data = TypeAssert\matches_type_structure(
+    type_alias_structure(TTypeAtPosOutput::class),
+    $untyped_data,
+  );
+
+  if (Shapes::keyExists($data, 'full_type')) {
+    if ($data['full_type']['kind'] === 'primitive') {
```
fredemmott: not a requirement, but altenratively:

```
if (($data['full_type']['kind'] ?? null) === 'primitive')
```

src/find_type_for_node.php
```diff
@@ -0,0 +1,78 @@
...
+      return $data['full_type']['name'] ?? null;
+    }
+    return $data['full_type']['kind'];
+  }
+  ;
```
fredemmott: delete

src/find_type_for_node.php
```diff
@@ -0,0 +1,78 @@
...
+    );
+  }
+  ;
+
+  $data = TypeAssert\matches_type_structure(
+    type_alias_structure(TTypeAtPosOutput::class),
+    $untyped_data,
+  );
+
+  if (Shapes::keyExists($data, 'full_type')) {
```
fredemmott: We like to reduce nesting - if you don't take the `??` approach:

```Hack
if (!Shapes::keyExists($data, 'full_type')) {
  return null;
}
```

tests/fixtures/migrations/hsl.php.expect
```diff
@@ -0,0 +1,127 @@
...
+  echo "\nucwords: ".Str\capitalize_words('hello world');
+  echo "\nucfirst: ".Str\capitalize("hello");
+  echo "\nstrtolower: ".Str\lowercase("HELLO");
+  echo "\nstrtoupper: ".Str\uppercase("hello");
+  echo "\nstr_replace: ".Str\replace('foobarbaz', 'foo', 'bar');
+  echo "\nstr_replace: ".
+    Str\replace_every('a test example with arrays', Dict\associate(['a', 'b'], ['x', 'y']));
+  $count = 0;
+  // this case is not handled in HSL
+  echo "\nstr_replace: ".str_replace('foo', 'bar', 'foobarbaz', &$count);
```
fredemmott: Is this because of the `&count` ? If so, can you put that in the comment

tests/fixtures/migrations/hsl.php.expect
```diff
@@ -0,0 +1,127 @@
...
+  echo "\nstr_ireplace: ".Str\replace_ci('foobarbaz', 'foo', 'bar');
+  // this case is also not handled
```
fredemmott: also extend this comment please, and the others below

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+  LiteralExpression,
+  ExpressionStatement,
+  OctalLiteralToken,
+  CommaToken,
+  WhiteSpace,
+};
+
+use function Facebook\HHAST\__Private\find_type_for_node;
+
+enum HSL_NAMESPACE: string as string {
```
fredemmott: is the `as string` required?

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+  C = 'C';
+  Dict = 'Dict';
+  Keyset = 'Keyset';
+  Math = 'Math';
+  Str = 'Str';
+  Vec = 'Vec';
+}
+
+type hsl_replacement_config = shape(
```
fredemmott: `UpperCamelCase` like class names: #112

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+
+use function Facebook\HHAST\__Private\find_type_for_node;
+
+enum HSL_NAMESPACE: string as string {
+  C = 'C';
+  Dict = 'Dict';
+  Keyset = 'Keyset';
+  Math = 'Math';
+  Str = 'Str';
+  Vec = 'Vec';
```
fredemmott: uppercase names (#113)

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+    ];
+
+
+  <<__Override>>
+  public function migrateFile(string $path, EditableNode $root): EditableNode {
+    // find all the function calls
+    $nodes = $root->getDescendantsOfType(FunctionCallExpression::class);
+
+    // keep track of any HSL namespaces we may have to add to the top of the file
+    $found_namespaces = dict[];
```
fredemmott: Why a dict rather than a `keyset` ?

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+      }
+    }
+
+    if (C\count($found_namespaces) === 0) {
+      return $root;
+    }
+
+    // add "use namespace" declarations at the top if they aren't already present
+    $declarations =
+      vec($root->getDescendantsOfType(INamespaceUseDeclaration::class));
```
fredemmott: This is ok for the common case, but:

```Hack
namespace Foo {
}

namespace Bar {
  use namespace HH\Lib\Vec;
}
```

... is also valid

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+    list($hsl_declarations, $suffixes) =
+      $this->findUseDeclarations($declarations);
```
fredemmott: Might want to look at ./__Private/Resolution/get_uses_directly_in_scope.php
./__Private/Resolution/get_current_uses.php
./__Private/Resolution/get_current_namespace.php

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+
+    $count_before = C\count(Vec\unique($suffixes));
```
fredemmott: efficiency: replace `Vec\unique($foo)` with `keyset($foo)` here: `Vec\unique($x)` is basically `vec(keyset($x))`

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+
+    // add new suffixes to the current list of suffixes
+    $suffixes = Vec\concat($suffixes, Vec\keys($found_namespaces))
+      |> Vec\unique($$);
```
fredemmott: would a keyset generally be more appropriate throughout?

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+
+    // added any new suffixes?
+    if (C\count($suffixes) !== $count_before) {
```
fredemmott: Reduce indentation: if ===, return early

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+        return (int)$expr->getText();
+      }
+
+      return null;
+    } elseif (
+      $node instanceof PrefixUnaryExpression &&
+      $node->getOperator() instanceof MinusToken
+    ) {
+      $val = $this->resolveIntegerArgument($node->getOperand());
+      return $val !== null ? -1 * $val : null;
```
fredemmott: Please add brackets - this is fine, but `? :`'s weirdness in PHP makes me nervous :p

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+      foreach ($argument_order as $index) {
+        $new_items[] = $items[$index];
+      }
+
+      $new_argument_list = vec[];
+
+      foreach ($arguments as $i => $argument) {
+        invariant($argument instanceof ListItem, 'expected ListItem');
+        $new_argument_list[] =
+          $argument->replace($argument->getItem(), $new_items[(int)$i]);
```
fredemmott: Is there a test covering this, both with and without trailing commas?

e.g, we can't turn `foo(1, 2, 3)` into `bar(3 1, 2,)` ?

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+    $suffixes = vec[];
+    foreach ($declarations as $decl) {
+
+      // we only care about "use namespace" directives
+      if (!($decl->getKind() instanceof NamespaceToken)) {
+        continue;
+      }
+
+      if ($decl instanceof NamespaceGroupUseDeclaration) {
+
```
fredemmott: delete

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+        $clauses = vec(
+          $decl->getClauses()->getDescendantsOfType(NamespaceUseClause::class),
+        );
+        foreach ($clauses as $clause) {
+          $name = $clause->getName();
+          if ($name instanceof NameToken) {
+            $suffixes[] = $name->getText();
+          }
+        }
```
fredemmott: This is fine - but alternatively (up to you):

```Hack
$suffixes = $clauses
  |> Vec\map($$, $c ==> { $n = $c->getName(); return $n instanceof NameToken ? $n->getText(): null; })
  |> Vec\filter_nulls($$)
  |> Vec\concat($suffixes, $$);
```

This is also another case where it feels like `keyset[]` is the right approach

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,646 @@
...
+    vec<string> $suffixes,
+  ): INamespaceUseDeclaration {
+    if (C\count($suffixes) > 1) {
+      // make a grouped use declaration
+      $ns = "{".Str\join($suffixes, ', ')."}";
+    } else {
+      // single namespace use declaration
+      $ns = C\firstx($suffixes);
+    }
+    return $this->nodeFromCode(
```
fredemmott: https://github.com/hhvm/hhast/issues/6 would be nice, but could be future work to clean this up, rather than a blocker.

src/__Private/MigrationCLI.php
```diff
@@ -21,6 +21,7 @@
   ImplicitShapeSubtypesMigration,
   OptionalShapeFieldsMigration,
   NamespaceFallbackMigration,
+  HSLMigration
```
fredemmott: Trailing comma

src/find_type_for_node.php
ssandler: Resolved! I had to make some changes due to the escaping. It was not possible to do input redirection due to the escaping built into `execute_async`, and I was having trouble getting `hh_client` to not complain when passed the path to the file (`hh_client --type-at-pos /tmp/file:line:offset` would complain about hh_server being stopped).

Turns out it wanted a path to a directory as a second arg, adding that allowed me to pass the file path and the containing directory, skipping input redirection. i.e. i.e. `hh_client --type-at-pos /tmp/file:line:offset /tmp`

src/find_type_for_node.php
ssandler: Thanks, took this suggestion

src/find_type_for_node.php
ssandler: makes sense. Used the `??` approach

tests/fixtures/migrations/hsl.php.expect
ssandler: Yes, commented.

tests/fixtures/migrations/hsl.php.expect
ssandler: Done

src/Migrations/HSLMigration.php
ssandler: Nope! I removed it and put `coerce()` in a few other places.

src/Migrations/HSLMigration.php
ssandler: Good point, fixed! I remembered keysets could only contain strings, and didn't think about the fact that the enum was ok to use in a keyset. This simplified a lot of logic.

src/Migrations/HSLMigration.php
ssandler: Should be addressed in https://github.com/hhvm/hhast/pull/108/commits/ba302ff77b6cea2ba4aae34fe7710048400071d9

I cleaned up the logic, using `insertBefore` to skip over namespace declarations, and also handled both kinds of namespace declarations (with and without body).

I didn't do anything special to handle two block namespace declarations in the same file, though. It will always insert it inside the first one.

src/Migrations/HSLMigration.php
ssandler: Any specific changes you're looking for here? I did crib off of these functions a bit, but I don't think I can use any of them directly?

src/Migrations/HSLMigration.php
ssandler: Added a test for this (the commas stay exactly where they are)

src/Migrations/HSLMigration.php
ssandler: That would be awesome (and as I just learned, it'd be good to be able to do that with insertBefore and insertAfter too). I'd feel much better about using a code string if it were part of the library :-).

Let's address separately though since this PR is getting large.

src/Migrations/HSLMigration.php
ssandler: I used this suggestion, thanks!

ssandler:
I believe all comments are addressed. I also added tests for the in and expect files having the same output.

## Running the migrator on hhast itself

As suggested outside of this review, I also ran the migration on hhast itself with `bin/hhast-migrate --hsl src`, which identified the issue with leading slashes that I handled in  https://github.com/hhvm/hhast/pull/108/commits/e3e16c6be696aead033c4825af4f839d0528102f. I verified tests pass on after running that migration. I can commit the results of that to this PR if you'd like.

```
SFO-M-SSANDLER01:hhast ssandler$ bin/hhast-migrate --hsl src
SFO-M-SSANDLER01:hhast ssandler$ git status
On branch ssandler_hsl_migration
Your branch is ahead of 'origin/master' by 23 commits.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   src/Linters/DontAwaitInALoopLinter.php
	modified:   src/__Private/SubprocessException.php
	modified:   src/__Private/codegen/CodegenSyntax.php
	modified:   src/nodes/EditableNode.php
	modified:   src/nodes/EditableToken.php
	modified:   src/nodes/EditableTrivia.php
	modified:   src/resolve_type.php

no changes added to commit (use "git add" and/or "git commit -a")
SFO-M-SSANDLER01:hhast ssandler$ vendor/bin/phpunit 
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

.....................................................I.........  63 / 295 ( 21%)
............................................................... 126 / 295 ( 42%)
............................................................... 189 / 295 ( 64%)
............................................................... 252 / 295 ( 85%)
...........................................                     295 / 295 (100%)

Time: 1.49 minutes, Memory: 64.06MB

OK, but incomplete, skipped, or risky tests!
Tests: 295, Assertions: 375, Incomplete: 1.
```

tests/MigrationsTest.php
```diff
@@ -158,4 +150,17 @@ classname<Migrations\BaseMigration> $migration,
...
     );
   }
+
+  public function testHslMigrationReturnsSameResults(): void {
+    $base_path = __DIR__.'/fixtures/migrations/hsl.php';
+    $phpstdlib_results =
+      \HH\Asio\join(HHAST\__Private\execute_async('hhvm', $base_path.'.in'));
+    $hsl_results = \HH\Asio\join(
+      HHAST\__Private\execute_async('hhvm', $base_path.'.expect'),
+    );
```
fredemmott: This should be a single HH\Asio\join, around a HH\Lib\Tuple\from_async

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+  C = 'C';
+  DICT = 'Dict';
+  KEYSET = 'Keyset';
+  MATH = 'Math';
+  STR = 'Str';
+  VEC = 'Vec';
+}
+
+type HslReplacementConfig = shape(
+  // \HH\Lib\* namespace the replacement function lives in
```
fredemmott: These should be `/** */` comments (supported by `hh-apidocs`)

src/Migrations/HSLMigration.php
fredemmott: Nope, just wanted to make sure you were aware of them, and that this needed doing here

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+
+    // remove any current use statements for HH\Lib\* namespaces, we'll group them together
+    $root = $root->removeWhere(
+      ($node, $parents) ==> C\contains($hsl_declarations, $node),
+    );
+
+    // build a possibly grouped namespace use declaration
+    $new_namespace_use_declaration = $this->buildUseDeclaration($suffixes);
+
+    // insert the nmew node: skip the <?hh sigil and namespace declaration if present,
```
fredemmott: nmew

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+        // insert the use statement inside the braces, before the first child
+        invariant($body instanceof NamespaceBody, 'expected NamespaceBody');
+        $first_child = $body->getDeclarationsx()->getChildren() |> C\firstx($$);
+        return
+          $root->insertBefore($first_child, $new_namespace_use_declaration);
+      }
+      return $root->insertBefore($child, $new_namespace_use_declaration);
+    }
+
+    invariant(false, 'should not fail to insert new node');
```
fredemmott: invariant_violation

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+        return (int)$expr->getText();
+      }
+
+      // a literal 0 shows as octal
+      if ($expr instanceof OctalLiteralToken) {
+        return (int)$expr->getText();
+      }
+
+      return null;
+    } elseif (
```
fredemmott: - space between `else` and `if`
- given the if block above returns, this could just be an `if`

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+    inout keyset<HslNamespace> $found_namespaces,
+  ): ?FunctionCallExpression {
+    $argument_list = $node->getArgumentList();
+    invariant($argument_list !== null, 'Function must have arguments');
+    $arguments = $argument_list->getChildren();
+    $new_argument_list = $argument_list;
+
+    $items = Vec\map(
+      $arguments,
+      ($argument) ==> {
```
fredemmott: parens uneeccessary

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+      return null;
+    }
+
+    // implode argument order is ambiguous
+    // when converting to join, check to make sure the second element is a string
+    // if the arguments are in the wrong order, reverse them
+    // if neither arg is a string, hh_client should complain so we just leave it as is
+    $fn_name = $this->getFunctionName($node);
+    if ($fn_name === 'Str\\join') {
+      $type = \HH\Asio\join(find_type_for_node_async($root, $items[1], $path));
```
fredemmott: make this function async, await result

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+      if ($type === 'string') {
+        $argument_order = Vec\reverse($argument_order ?? vec[]);
+      }
+    } elseif ($fn_name === 'Str\\replace' || $fn_name === 'Str\\replace_ci') {
```
fredemmott: else if

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+      // str_replace and str_ireplace have two modes:
+      // string for search/replace args means replacing a single pattern
+      // arrays mean replacing a set of patterns, which we should rewrite as Str\replace_every
+      $type = \HH\Asio\join(find_type_for_node_async($root, $items[0], $path));
```
fredemmott: await

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,684 @@
...
+  ): EditableNode {
+    $parents = null;
+    $found = false;
+    $stack = $root->findWithParents($it ==> $it === $node);
+    invariant(!C\is_empty($stack), 'did not find node in root');
+    invariant(C\lastx($stack) === $node, 'expected node at top of stack');
+    $stack_count = C\count($stack);
+    $parent = $stack[$stack_count - 2];
+
+    if ($parent instanceof BinaryExpression) {
```
fredemmott: this also looks like a good candidate for negate the check, return early, unindent the rest

ssandler:
Just pushed fixes to all of those!

src/Migrations/HSLMigration.php
```diff
@@ -0,0 +1,711 @@
...
+      'count' => shape(
+        'ns' => HslNamespace::C,
+        'name' => 'count',
+        'argument_order' => vec[0],
+      ),
+    ];
+
+
+  <<__Override>>
+  public function migrateFile(string $path, EditableNode $root): EditableNode {
```
fredemmott: heh, I should change this to be an async function in the next version. THis is fine :)

PR #117, opened 2018-08-08
Ssandler hsl migration

ssandler:
I ran the migration on this repo and ensured the tests still pass.


ssandler:
It would be a little bit nicer to:
- avoid the trailing newline if the following declaration is a namespace use declaration to avoid the extra newline
- if we removed any namespace use declarations that are HSL declarations, put the new one in the same place

ssandler:
I made the change to remove the extra newline if the following declaration is a `use` statement and re-ran the migration, looks a bit nicer.

Issue #107, opened 2018-07-20
assert to expect migration: `use function`


fredemmott:
- do not insert `use function` if there is an existing `use function ....\expect`, even if it's a different one (HHAST has its' own one)
- do not insert `use function` if there aren't any calls to expect() in the file after the migration

PR #118, opened 2018-08-09
Do not insert use function under certain conditions

wlin57:
#107


src/Migrations/AssertToExpectMigration.php
```diff
@@ -71,13 +72,14 @@ private static function getExpectFunction(): NamespaceUseDeclaration {
   private function checkExpectFunctionPresent(
     NamespaceUseDeclaration $node,
   ): NamespaceUseDeclaration {
-    if ($this->expectPresent) {
+    if (!$this->expectPresent) {
```
fredemmott: 🎁

PR #120, opened 2018-08-13
Use only one bool

wlin57:
Don't need two booleans to determine whether we need to add `use function Facebook\FBExpect\expect;` 
`useExpectFunctionNeeded` starts out false and is only set to true if there are expect calls. It is set to false if the use statement already exists or if the migration adds it.


fredemmott:
restarted the travis job, that looked pretty bogus

fredemmott:
@wlin57 can you check this? It looks like hh_server is insta-crashing on this diff, at least on travis

PR #125, opened 2018-08-28
Faster relationship regeneration

fredemmott:
- use the new async entrypoints to the parser
- smaller wait
- higher limit: still not particularly intensive, but actually useful
now we have a smaller wait

This gets the time from > 20 minutes to ~ 5 minutes on a mac pro


fredemmott:
This also includes hackfmt

PR #128, opened 2018-09-13
Add PHPUnit => HackTest migration

fredemmott:
This should be rebased then merged, not squashed.

it's probably best to review each commit individually


fredemmott:
also, @wlin57 fyi :)

itshappening-cook

fredemmott:
This has now been used for:
- HHAST itself
- hh-clilib
- fbmarkdown

PR #131, opened 2018-09-17
Make LSP message flushing async, complete, and chunk in-order

### LINKED ISSUE #130: Make sure LSP writes are complete, and in-order
we use `AsyncPoll` to interleave multiple concurrent requests/responses; a response could conceivably be bigger than the buffer. We need to make sure that:
- we output the whole thing
- we output the entirety of one message before we start the next

This is likely to involve:
- updating to https://github.com/hhvm/hh-clilib/releases/tag/v1.2.1
- an output queue like this:

```Hack
final class OutputQueue {
  public function __construct(
    private OutputInterface $out,
  ) {
  }
  private ?Awaitable<void> $tail;

  public function enqueueOutput(string $str): Awaitable<void> {
    $tail = $this->tail;
    $next = async {
      await $tail;
      await $this->out->writeAsync($str);
    };
    $this->tail = $next;
    return $next;
  }
}
```

fredemmott:
fixes #130

`write()` might only write part of the data (it returns number of bytes written). `writeAsync()` (new) does the whole lot, but can interleave.

1. Convert all the client => server methods to async
2. Use writeAsync
3. Build a queue so that we don't get a chunk from one message
interleaved with a chunk from another

### DIFF
--- a/src/__Private/LSPImpl/Client.php
+++ b/src/__Private/LSPImpl/Client.php
@@ -15,15 +15,26 @@ use type Facebook\CLILib\ITerminal;
 use namespace HH\Lib\Str;
 
 final class Client extends LSPLib\Client {
+  private ?Awaitable<void> $pendingStdout;
+
   public function __construct(private ITerminal $terminal) {}
 
+  /** This is intentionally not an async function; it stores/directly creates
+   * waithandles to preserve ordering */
   <<__Override>>
-  protected function sendMessage(LSP\Message $message): void {
+  protected function sendMessageAsync(LSP\Message $message): Awaitable<void> {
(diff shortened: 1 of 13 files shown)


src/__Private/LSPImpl/Client.php
```diff
@@ -15,15 +15,26 @@
...
+    $next = async {
+      await $tail;
+      await $this->terminal
+        ->getStdout()
+        ->writeAsync(
+          Str\format("Content-Length: %d\r\n\r\n%s", Str\length($json), $json),
+        );
+    };
+    $this->pendingStdout = $next;
+    return $next;
```
fredemmott: ^ this bit is the fun part ; cc @jan, @kmeht

Issue #110, opened 2018-08-06
Add linter banning empty statements


fredemmott:
e.g. `;` as an entire statement.

for loops are a possible exception, and should be tested.

ryangreenberg:
I think this can be closed after #139 and #145

PR #139, opened 2018-11-12
Add linter banning empty statements

ryangreenberg:
This linter addresses #110.

I'd like input on two areas in particular:

1. The language of the error. Is there something more helpful/friendly to say than "Delete this empty statement"?
2. Whitespace management

I was unsure how much whitespace management to do in this particular linter. I think there are a few options:

- Do nothing (remove the entire expression, including leading and trailing whitespace)
- Retain all whitespace (just remove the semicolon, current implementation)
- Retain only newline whitespace

I started with do nothing. My thinking was to punt and let `hackfmt` handle reformatting. But the autofix output looked pretty bad, so I tried to retain all the leading/trailing whitespace. It looks fine but can cause other errors (e.g. trailing whitespace). Retaining only newlines feels like it _might_ be a good compromise but it might just be more finicky.

As noted in #110 this linter permits use of semicolons in `for` loops like `for (;;)`.

### DIFF
--- a/src/Linters/NoEmptyStatementsLinter.php
+++ b/src/Linters/NoEmptyStatementsLinter.php
@@ -10,11 +10,7 @@
 
 namespace Facebook\HHAST\Linters;
 
-use type Facebook\HHAST\{
-  ExpressionStatement,
-  EditableNode,
-  EditableList,
-};
+use type Facebook\HHAST\{ExpressionStatement, EditableNode, EditableList};
 
 final class NoEmptyStatementsLinter
   extends AutoFixingASTLinter<ExpressionStatement> {
@@ -36,11 +32,7 @@ final class NoEmptyStatementsLinter
 
     $expr = $stmt->getExpression();
     if ($expr === null) {
-          return new ASTLintError(
-      $this,
-      'This statement is empty',
-      $stmt,
-    );
+      return new ASTLintError($this, 'This statement is empty', $stmt);
     }
 
     return null;
@@ -52,6 +44,7 @@ final class NoEmptyStatementsLinter
     $leading = $semicolon->getLeading();
     $trailing = $semicolon->getTrailing();
 
-    return EditableList::concat($semicolon->getLeading(), $semicolon->getTrailing());
+    return
+      EditableList::concat($semicolon->getLeading(), $semicolon->getTrailing());
   }
 }
--- a/src/Linters/NoEmptyStatementsLinter.php
+++ b/src/Linters/NoEmptyStatementsLinter.php
@@ -42,8 +42,8 @@ final class NoEmptyStatementsLinter
   }
 
   <<__Override>>
-  public function getFixedNode(ExpressionStatement $expr): EditableNode {
-    $semicolon = $expr->getSemicolonx();
+  public function getFixedNode(ExpressionStatement $stmt): EditableNode {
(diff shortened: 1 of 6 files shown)


src/Linters/NoEmptyStatementsLinter.php
```diff
@@ -0,0 +1,52 @@
...
+
+  <<__Override>>
+  public function getLintErrorForNode(
+    ExpressionStatement $stmt,
+    vec<EditableNode> $_parents,
+  ): ?ASTLintError<ExpressionStatement> {
+
+    $expr = $stmt->getExpression();
+    if ($expr === null) {
+          return new ASTLintError(
```
fredemmott: over-indented; below is under-indented

src/Linters/NoEmptyStatementsLinter.php
```diff
@@ -0,0 +1,52 @@
...
+      $this,
+      'Delete this empty statement',
```
fredemmott: Sorry if I've not followed this completely, but I aim for the errors to describe the problem, not the fix - maybe 'this statement is empty' or just 'empty statement'?

The fix message (shown by IDEs) can be specified separately

ryangreenberg:
Thanks for the review, @fredemmott. I tweaked the error/fix language, added the additional test cases, and ran the src and test file through `hackfmt`.

ryangreenberg:
Thanks for the review, @fredemmott. I tweaked the error/fix language, added the additional test cases, and ran the src and test file through `hackfmt`.

src/Linters/NoEmptyStatementsLinter.php
ryangreenberg: Fixed by using `hackfmt` 😅

src/Linters/NoEmptyStatementsLinter.php
ryangreenberg: Now: "This statement is empty" and fix message is "Remove statement".

tests/fixtures/NoEmptyStatementsLinter/empty_statements.php.autofix.expect
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+function foo(): void {
+  some_function();
+  
+
+  // leading comment
+  
+
+   // trailing comment
```
fredemmott: sorry for the slow review; are you able to fix this indentation reasonably?

ryangreenberg: No worries!

I'm not sure what the best approach is here so I'm very open to input. What's shown here is the result of retaining all the leading and trailing whitespace around the statement (option #2 in the initial PR description).

I can't think of something simple that gives you good output for all of these scenarios:

```Hack
function whitespace_options(): void {
  ; // trailing comment
  ;   // trailing comment
  ;
}
```

Do you have any ideas in mind?

fredemmott: Let's leave this as-is; several linters/migrators are now handling this in their own way - it's probably best fixed by adding a generic "fix the whitespace for this trivia" helper (`hackfmt`-style, not neccessarily trying to preserve it) and replacing all the existing things with it.

PR #140, opened 2018-11-15
Add linter to convert PHP anonymous functions to lambdas

ryangreenberg:
This change adds an autofixing linter that converts [PHP anonymous functions](http://php.net/manual/en/functions.anonymous.php) to [Hack lambdas](https://docs.hhvm.com/hack/lambdas/introduction). It does not fix anonymous functions that use variables by reference which have no equivalent lambda.

In my implementation of `getFixedNode` I am not sure if there is a better approach to copying the child nodes from the anonymous function to the lambda expression. I use `?? Missing()` extensively; any suggestions for a better approach are very welcome.

A future improvement which I did not implement here would be to examine the body of the anonymous function and, if it returns the result of a single expression, rewrite body to drop the curly braces, return keyword, and semicolon.

Specifically:

```Hack
$fn = function($i) { return $i * $i; };

// currently becomes
$fn = ($i) ==> { return $i * $i; };

// an improvement would be
$fn = ($i) ==> $i * $i;
```

### DIFF
--- a/src/Linters/PreferLambdasLinter.php
+++ b/src/Linters/PreferLambdasLinter.php
@@ -65,25 +65,31 @@ final class PreferLambdasLinter extends AutoFixingASTLinter<AnonymousFunction> {
 
   <<__Override>>
   public function getFixedNode(AnonymousFunction $node): ?EditableNode {
-    $attribute_spec = $node->getAttributeSpecUNTYPED();
-    $async = $node->getAsyncKeywordUNTYPED();
-    $coroutine = $node->getCoroutineKeywordUNTYPED();
+    $attribute_spec = $node->getAttributeSpec();
+    $async = $node->getAsyncKeyword();
+    $coroutine = $node->getCoroutineKeyword();
+    $parameters = $node->getParameters();
     $left_paren =
       new LeftParenToken($node->getFunctionKeyword()->getLeading(), Missing());
+    $right_paren = $node->getRightParen();
+    $colon = $node->getColon();
+    $type = $node->getType();
+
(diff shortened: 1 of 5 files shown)


src/Linters/PreferLambdasLinter.php
```diff
@@ -0,0 +1,98 @@
...
+      !C\is_empty(
+        $use_expr->getDescendantsWhere(
+          ($node, $_) ==> $node is PrefixUnaryExpression &&
+            $node->getOperator() is AmpersandToken,
+        ),
+      );
+
+    if ($uses_references) {
+      return null;
+    } else {
```
fredemmott: `else` unneeded because `if` body is noreturn; move else body to function body and unindent for readability

src/Linters/PreferLambdasLinter.php
```diff
@@ -0,0 +1,98 @@
...
+    $signature = new LambdaSignature(
+      $left_paren,
+      $parameters ?? Missing(),
+      $right_paren ?? Missing(),
+      $colon ?? Missing(),
+      $type ?? Missing(),
+    );
+
+    $arrow = new EqualEqualGreaterThanToken(Missing(), new WhiteSpace(' '));
+		$body = $node->getBody();
```
fredemmott: formatting

fredemmott:
(also waiting for others to give feedback on potential edge cases - none come to mind, but being a little paranoid)

fredemmott:
Thanks; can't find any cases, so I guess let's ship it and see what happens :) Worst case, it can easily be disabled in the config file

PR #143, opened 2018-12-03
Fix generated @return docblock

muglug:
The current codegen generates `@returns` docblocks. I know you're not following PSR-*, but I also don't see reference to `@returns` vs `@return` in other documentation, so I assume it's a typo.

This also replaces `Missing` with `null` in the docblock where the method has an explicit `isMissing` check.


fredemmott:
Thanks!

PR #145, opened 2018-12-03
Improve NoEmptyStatementLinter to detect more types of empty statements (Fixes #142)

### LINKED ISSUE #142: Improve NoEmptyStatementLinter to detect more types of empty statements
As discussed in the hacklang Slack, it would be nice for the `NoEmptyStatementLinter` to flag expressions that are semantically empty (expressions that have no effect on the code).

Here are some examples of things we can flag:

```Hack
function more_empty_exprs(): void {
  // Obviously empty
  $a;
  $a.$b;
  $a + $b;
  "a"."b";
  $a.$b.$c;

  // Less clear
  // The concatenation is useless, but maybe fn() has side effects
  // Probably worth flagging as a problem; not sure if we should
  // avoid an autofix
  $a . fn();

  // Not empty—should not be considered empty statements
  side_effect();
  $counter++;
  $i = 5;
  $i += 1;
  $i--;
}
```

My initial thinking is to flag any `ExpressionStatement` where the expression node is one of the following:

- `VariableExpression`
- `BinaryExpression` where the operator is one of the non-assignment tokens (e.g. `PlusToken`, but not `PlusEqualsToken`)

How does this approach sound? Are there any other edge cases to consider?

Edit: This is, of course, more complicated than I initially thought. I went through all the possible values for `ExpressionStatement::getExpression` and I think there is a long list of unambiguous cases to flag, plus `BinaryExpression`, where we should flag most expressions without an assignment operator, and a couple other cases. I'll try running some sample code on a large codebase and report back.

ryangreenberg:
This change flags a wide variety of semantically empty statements (statements that have no effect on the program). For example:

```Hack
'a'.'b';
1 + 2;
$a === $b;
```

These statements do nothing without an assignment or a function call.

At first I changed the autofix to _not_ suggest a correction when the statement included children that may have side effects (function calls, `as` expressions which can throw, etc.). After running this linter on one of Slack's Hack codebases I changed my mind: I don't think it's a good idea to try to autofix semantically empty statements at all. Almost all of the statements found by this linter represent bugs which require manual intervention to address. Here are some (redacted) examples:

```
$dict[$key] == "abc"; // should _probably_ be "=" for assignment
my_function()['key']; // the ['key'] is useless but the right fix is non-obvious.
$dict[$key]; // maybe a mistake, maybe should be an assignment?
var_export(my_function())."\n"; // probably want the newline inside?
putenv('VAR=').$val; // bug--should be inside the function call
```

It even identified a (minor) bug in [an external library](https://github.com/stripe/stripe-php/blob/913b20c782b7d27accfe7fa1a88c44b3ee27e998/lib/Card.php#L102)!

## Notes

- I didn't want to import dozens of types from Facebook\HHAST, so I did `use namespace`. Hope this is OK.
- I'm not sure how important it is for this linter to support `do_something() || die("failed");` which it currently flags as empty. My two cents is that this is not that common and easy enough to rewrite or `HHAST_IGNORE_ERROR`.
- I found it easier to include all the no-op cases in the fixture input rather than writing them in `getCleanExamples`.


ryangreenberg:
Looking at the build failure

ryangreenberg:
Found another edge case to consider:

```Hack
fn1() && fn2();
```

Since we already consider function calls non-empty by default, a simple somewhat principled fix would be to consider binary expressions non-empty if both sides are a function call.

PR #152, opened 2018-12-26
Limit descendants to namespace name

muglug:
This prevents an error with

```php
namespace Name\Space {
    class A {
    }
    class B {
    }
}
```

where the given namespace is `Name\Space\A\B`

Happy to make a test if you want.

### DIFF
--- a/src/nodes/NamespaceDeclaration.hack
+++ b/src/nodes/NamespaceDeclaration.hack
@@ -18,7 +18,11 @@ final class NamespaceDeclaration extends NamespaceDeclarationGeneratedBase {
       return $name->getText();
     }
 
-    return $this->getDescendantsOfType(NameToken::class)
+    if ($name === null) {
+      return '';
+    }
+
+    return $name->getDescendantsOfType(NameToken::class)
       |> Vec\map($$, $t ==> $t->getText())
       |> Str\join($$, "\\");
   }
--- a/tests/ResolutionTest.hack
+++ b/tests/ResolutionTest.hack
@@ -25,9 +25,9 @@ final class ResolutionTest extends TestCase {
     $code = '<?hh namespace MyNS\\SubNS { class Foo {}; class Bar {} }';
     $ast = from_code($code);
     $namespaces = $ast->getDescendantsOfType(NamespaceDeclaration::class);
-    $class_names = $ast->getDescendantsOfType(NamespaceDeclaration::class)
+    $namespace_names = $ast->getDescendantsOfType(NamespaceDeclaration::class)
       |> Vec\map($$, $namespace ==> $namespace->getQualifiedNameAsString());
-    expect($class_names)->toBeSame(vec[
+    expect($namespace_names)->toBeSame(vec[
       "MyNS\\SubNS",
     ]);
   }
--- a/tests/ResolutionTest.hack
+++ b/tests/ResolutionTest.hack
@@ -13,11 +13,23 @@ namespace Facebook\HHAST;
 use function Facebook\FBExpect\expect;
 use namespace Facebook\HHAST\__Private\Resolution;
 use type Facebook\HackTest\DataProvider;
+use namespace HH\Lib\Vec;
 
 final class ResolutionTest extends TestCase {
   public function testWithoutNamespaces(): void {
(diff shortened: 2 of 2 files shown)


azjezz:
@muglug tests are great c:

src/nodes/NamespaceDeclaration.php
```diff
@@ -18,8 +18,12 @@ public function getQualifiedNameAsString(): string {
     if ($name instanceof NameToken) {
       return $name->getText();
     }
+    
+    if (!$name) {
```
azjezz: `if(null === $name)` or `if(!($name is nonnull))`

muglug: Thanks! I'm not much of a fan of Yoda conditionals, so I've updated with `$name === null`.

src/nodes/NamespaceDeclaration.php
```diff
@@ -18,8 +18,12 @@ public function getQualifiedNameAsString(): string {
     if ($name instanceof NameToken) {
       return $name->getText();
     }
+    
+    if ($name === null) {
+      return '';
```
azjezz: shouldn't this be : 
```hack
     return $this->getDescendantsOfType(NameToken::class)
       |> Vec\map($$, $t ==> $t->getText())
       |> Str\join($$, "\\");
```
? in that case, instead of write that code, you can do `$name = $this` without `return` which should do the same

muglug: No - I think that would break with

```php
namespace {
  class A {}
  class B {}
}
```

Sadly I haven't time to add test cases at the moment, but will be more free in first few days of January.

fredemmott:
Hey, sorry for the delay - I got the flu which is getting me a bit of a back log. I hope to get this merged this week or next.

fredemmott:
This looks good, but does need a test case adding - thanks

fredemmott:
@muglug : sorry, could you add a test? I tried this, based on your example, but it passes:

```
diff --git a/tests/ResolutionTest.hack b/tests/ResolutionTest.hack
index b572df3..f1cba13 100644
--- a/tests/ResolutionTest.hack
+++ b/tests/ResolutionTest.hack
@@ -13,6 +13,7 @@ namespace Facebook\HHAST;
 use function Facebook\FBExpect\expect;
 use namespace Facebook\HHAST\__Private\Resolution;
 use type Facebook\HackTest\DataProvider;
+use namespace HH\Lib\Vec;

 final class ResolutionTest extends TestCase {
   public function testWithoutNamespaces(): void {
@@ -36,6 +37,21 @@ final class ResolutionTest extends TestCase {
       ->toBeSame('MyNS\\SubNS');
   }

+  public function testMultipleClassesInNamespaceBlock(): void {
+    $code = '<?hh namespace MyNS\\SubNS { class Foo {}; class Bar {} }';
+    $ast = from_code($code);
+    $class_names =
+      $ast->getDescendantsOfType(ClassishDeclaration::class)
+      |> Vec\map($$, $class ==> {
+        $parents = $ast->findWithParents($x ==> $x === $class);
+        return Resolution\get_current_namespace($class, $parents)."\\".($class->getName() as EditableToken)->getText();
+      });
+    expect($class_names)->toBeSame(vec[
+      "MyNS\\SubNS\\Foo",
+      "MyNS\\SubNS\\Bar",
+    ]);
+  }
+
   public function testWithUnrelatedNamespaceBlock(): void {
     list($node, $parents) = self::getNodeAndParents(
       '<?hh namespace \\MyNS { } namespace { class Foo {} }'
```

muglug:
I'll have a look, sorry!

muglug:
I've added a test that calls `NamespaceDeclaration::getQualifiedNameAsString` which is the thing (I think) that's buggy. Sorry to force push - was simplest given mass renaming.

muglug:
Without the fix the test outputs
```
Failed asserting that vec [
  'MyNS\\SubNS\\Foo\\Bar',
] is the same as vec [
  'MyNS\\SubNS',
]
```

muglug:
> the name of the namespace itself is the problem?

Yup!

PR #154, opened 2019-02-12
migrate to `.hack` file extension

azjezz:
hh-to-hack migrate src  && hh-to-hack migrate tests && hh-to-hack migrate codegen


azjezz:
cc @fredemmott o.o

fredemmott:
Separately done as part of #155

PR #157, opened 2019-03-06
Add `<?hh` -> `<?hh // partial` migration

fredemmott:
... but use strict mode instead please :)


src/__Private/MigrationCLI.hack
```diff
@@ -167,14 +182,14 @@ class MigrationCLI extends CLIWithRequiredArguments {
...
-    if(\HHVM_VERSION_ID >= 33100) {
+    if (\HHVM_VERSION_ID >= 33100) {
       $options[] = CLIOptions\flag(
         () ==> {
           $this->migrations[] = AwaitPrecedenceMigration::class;
         },
         'Parenthesize await operands that would break if precedence of await changes',
         '--await-precedence',
-        );
+      );
```
fredemmott: hackfmt

PR #161, opened 2019-04-27
[25% less time] Pass references to source code to elements when const…

fredemmott:
…ructing

This heavily optimizes `getCode()` for unmodified nodes. Source
references are used instead of raw strings because raw strings end up
exponential in memory usage

For example:

`foo($a, $b)`

- function call would store `foo($a, $b)`
- param list would store `($a, $b)`
- vars would store `$a`

This gets much worse when you think about functions in while loops in
methods in classes in namespaces.

Given an extremely large file (all of xhp-lib core cat'ed in to one
file), this takes the lint time on my laptop down from 12s to 9s.


fredemmott:
@bradleyhd are you able to review this? :)

fredemmott:
Current results for this branch /with a warm jit/ (will merge without squashing after review):

# Before

```
real	0m2.150s
user	0m0.625s
sys	0m0.105s
```

# After

```
real	0m1.553s
user	0m0.635s
sys	0m0.107s
```

bradleyhd:
Sorry I just saw this, was on vacation. I don't work at Prev Co anymore but happy to keep looking at PRs so keep 'em coming!

PR #181, opened 2019-06-06
Implement GroupUseStatementAlphabetizationLinter

### LINKED ISSUE #115: Add linter for group use statement alphabeticization
eg

```Hack
use namespace Foo\{
  C,
  A,
  B,
}
```

should be rewritten to

```Hack
use namespace Foo\{
  A,
  B,
  C,
}
```

aloiret:
Fixes #115


aloiret:
Doesn't work with multiple NameTokens in NamespaceUseClause e.g.

```hack
use namespace HH\{Lib\Vec, Lib\Str};
```

working on it.

PR #183, opened 2019-06-07
Wrap NameToken/QualifiedName in List<ListItem<IExpression>>

### LINKED ISSUE #180: List<IExpression>/ListItem<IExpression> do not appropriately wrap NameToken/QualifiedNameToken
A `NameExpression` should be returned, wrapping these.

fredemmott:
fixes #180

Previously would wrap in accessors; instead wrap when converting the
JSON to objects, passing through type information.

- fixed several unit test that were checking invalid Hack code, as this
moves "expected expression" errors to an earlier phase
- added runtime tests for `IExpression` and `List<ListItem<IExpression>>`
- while I was there, removed `as dict<_, _>` given poor
non-repo-auth-mode performance

### DIFF
--- a/codegen/node_from_json.hack
+++ b/codegen/node_from_json.hack
@@ -1,24 +1,26 @@
 /**
  * This file is generated. Do not modify it manually!
  *
- * @generated SignedSource<<8f34279a4c1a6fd21bbc06653547c514>>
+ * @generated SignedSource<<5c6ffcc311022b149d0fb1337c916ebb>>
  */
 namespace Facebook\HHAST\__Private;
 use namespace Facebook\HHAST;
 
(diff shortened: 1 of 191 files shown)


fredemmott:
@hgoldstein @KendallHopkins I've done it this way instead of pre-processing the JSON given that HHAST is heavily used for as-you-type linting, so there's minimal benefit in caching pre-processed JSON

fredemmott:
> an edge needs to have a promotion. 

In the case of List<Item<T>> is the edge to the list (I.e do you rewrite lists, items, and item.item?) or the Item -> child edge?

fredemmott:
c366ffc

PR #186, opened 2019-06-10
Wrap type-like tokens as ITypeSpecifiers

### LINKED ISSUE #185: Add Specifier token promotion
Part of #179

- [x] `foo(): noreturn` has a function type of just `token:noreturn`; this is very strange: no other tokens are seen there without wrapping, and noreturn is valid inside a simple_type_specifier; in the case of `foo(): this` ,there's a simpletypespecifier; in `foo(): Awaitable<noreturn>` there is also a simple type specifier. This should probably be addressed inside Hack's parser to remove this special edge case (FB T45648628)
- [x] cases like type_constant.type_constant_left_type should return `ITypeSpecifier`, which means that `token:name`, `token:parent`, `token:self`, and `token:this` need wrapping

Could either wrap into a dedicated wrapper class (like is done for IExpression), or SimpleTypeSpecifier might be directly usable

fredemmott:
fixes #185

### DIFF
--- a/codegen/syntax/FunctionDeclarationHeader.hack
+++ b/codegen/syntax/FunctionDeclarationHeader.hack
@@ -1,7 +1,7 @@
 /**
  * This file is generated. Do not modify it manually!
  *
- * @generated SignedSource<<68a976e87d6302184ee2703cf306cc04>>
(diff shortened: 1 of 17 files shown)


fredemmott:
fixed by 2f67c09a69d3eb3065294b27b032f52792935500

PR #190, opened 2019-07-05
Add linter to detect unused variables

ryangreenberg:
This change adds the `UnusedVariableLinter` which detects unused variables in a function or method body. It is based on `UnusedParameterLinter` and has the same error message and autofix suggestion.

**Approach**
The strategy used here is to classify every variable expression in a function/method body as an assignment or a use. If the variable expression being considered does not appear in the list of uses, it's an error. The details of classification are documented at `UnusedVariableLinter::classifyVariables`:

```
   * An assignment is a variable expression that is one of the following:
   *
   * 1a. The left-hand side of a binary expression with an assignment operator
   *   that is not part of the index of a subscript expression.
   * 2. The key or value of a foreach clause
   * 3. The members of a list expression
```

**Caveats**
This linter assumed that no variables are used before they are assigned, which should be enforced separately by the typechecker. It may have false negatives (variables that are unused but not detected as such) either because of bugs or unreachable uses (e.g. `$a = 1; return 5; return $a;`).

**Performance**
`UnusedVariableLinter::classifyVariables` is probably slow because it needs the parents for the variable expression node. For small functions this won't matter much; for large functions I made some performance improvements in separate commits:
1. Use `MemoizableNode` to classify variables once per function/methods
2. Switch vecs to keysets

If these changes make sense, `MemoizableNode` probably belongs somewhere else.

### DIFF
--- a/src/Linters/UnusedVariableLinter.hack
+++ b/src/Linters/UnusedVariableLinter.hack
@@ -187,9 +187,6 @@ final class UnusedVariableLinter extends AutoFixingASTLinter {
     VariableExpression $var,
     vec<Node> $parents,
   ): bool {
-    // Whether we have encountered a $parent SubscriptExpression where SubscriptExpression::getReceiver === $var
-    $is_subscript_receiver = null;
-
     foreach (Vec\reverse($parents) as $parent) {
       if ($parent === $var) {
         continue;
@@ -212,11 +209,11 @@ final class UnusedVariableLinter extends AutoFixingASTLinter {
       // $a[$b] = '123'; // assignment of $a, use of $b
       // f($a[$b]);      // use of $a
       if ($parent is HHAST\SubscriptExpression) {
-        $is_subscript_receiver ??= $parent->getReceiver() === $var;
-        if ($is_subscript_receiver) {
-          continue;
-        } else {
+        $index = $parent->getIndex();
+        if ($index !== null && $index->isAncestorOf($var)) {
           return false;
+        } else {
+          continue;
         }
       }
 
--- a/src/Linters/UnusedVariableLinter.hack
+++ b/src/Linters/UnusedVariableLinter.hack
@@ -22,7 +22,7 @@ use type Facebook\HHAST\{
 };
 
 use namespace Facebook\HHAST;
-use namespace HH\Lib\{C, Str, Vec};
+use namespace HH\Lib\{C, Keyset, Str, Vec};
 
 final class UnusedVariableLinter extends AutoFixingASTLinter {
   const type TNode = VariableExpression;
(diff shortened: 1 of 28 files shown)


ryangreenberg:
There is a bug in this linter that needs to be fixed before it is ready to merge:

- correct handling of subscript expressions in non-assignment contexts. $a['b'] should count as a use of `$a`.

ryangreenberg:
I found one more edge case involving `EmbeddedBraceExpression` that complicates detection of variables used with the `${foo}` or `${foo['bar']}` forms, since `foo` is a NameToken inside the expression. I'm not sure whether to (a) fix this or (b) encourage people to fix it by adopting the `{$foo}` alternative.

ryangreenberg:
Another edge case that needs to be fixed:

```
foreach ($xs as $x) {
  $dict[$x['blah']] = 1;
}
```

incorrectly flags `$x` as unused

ryangreenberg:
Here's the current state of this PR:

- [x] Classify `$a['b']` as a use of `$a`
- [x] Classify `$c[$a['b']] = 'b'` as a use of `$a`
- [ ] Classify variable use in `EmbeddedBraceExpression` like "${a}" as a use of `$a` (pending decision about whether we want to do it because it's a little annoying; `{$a}` is already handled correctly.

lexidor:
A `{$var}` migration got put into HHAST. (you are most likely already aware)

fredemmott:
I expect `"${foo}"` to be removed from the language in the next few weeks, possibly today

fredemmott:
Thanks; can you add it to the 'ALL' linter group, and run it against hhast itself?

- if it doesn't find anything, please intentionally add something and make sure it fixes when ran from CLI
- if it does, make sure they're valid, and add the fixes to this PR as a separate commit

fredemmott:
It would also be good to run this against vendor/ and put what it finds and suggests in a gist

ryangreenberg:
> - [x] add it to the 'ALL' linter group,
> - [x] and run it against hhast itself?

I applied a bunch of fixes in a separate commit. I think they're all legitimate, including one typo. I made some judgment calls in, e.g. dropping `$key` from a `foreach` instead of changing it to `$_key`. Happy to go another way if you prefer.

Here's a gist with the results of running the linter on `vendor`: https://gist.github.com/ryangreenberg/d96c0f2fccacf7a0f14fa1f49d290b53

src/Linters/UnusedVariableLinter.hack
```diff
@@ -0,0 +1,289 @@
...
+
+  public function getFixedNode(VariableExpression $node): VariableExpression {
+    $expr = $node->getExpression();
+    if (!$expr is VariableToken) {
+      return $node;
+    }
+    return $node->withExpression(
+      $expr->withText('$_'.Str\strip_prefix($expr->getText(), '$')),
+    );
+  }
```
jjergus: Wouldn't it be a better fix to remove the whole assignment? (except for special cases like `list($_unused, $used) = ...`)

I'm slightly concerned that this suggest what is arguably the wrong fix in most cases, but we can discuss that later in a separate issue/PR.

I imagine that correctly removing the assignment could be much harder.

jjergus:
None of my suggestions above are very important so I merged this, but we can continue discussing here and create a new PR if we decide to change anything.

PR #194, opened 2019-07-12
Instanceof is migration

fredemmott:


fredemmott:
There's `--add-fixmes` ; that's probably appropriate for some projects but others will want manual inspection

PR #197, opened 2019-07-18
Add migration for top-level requires

fredemmott:
- if there is a valid `<<__EntryPoint>>` function, move the requires there inside a lambda
- otherwise ~~, add an easily greppable `/* HH_FIXME[1002] `~~ assume it's not in strict mode and leave it alone

### DIFF
--- a/bin/hhast-inspect
+++ b/bin/hhast-inspect
@@ -10,11 +10,13 @@
 
 namespace Facebook\HHAST\__Private;
 
-require_once(__DIR__.'/hhast-inspect.hack');
-
 // As this file does not have an extension, it is not typechecked. Delegate
 // to the typechecked one.
 <<__EntryPoint>>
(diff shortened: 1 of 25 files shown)


fredemmott:
Not safe.

```
$ cat strict.php
<?hh // strict

<<__EntryPoint>>
function main(): void {
  $foo = 'bar';
  \var_dump($foo);
  require_once('partial.php');
  \var_dump($foo);
}
$ cat partial.php
<?hh
$foo = 123;
$ hhvm strict.php
string(3) "bar"
int(123)
```

fredemmott:
Not safe, e.g.:

```
require_once('autoloader.php' /* so that `Bar` is available */);
class Foo extends Bar {}
<<__EntryPoint>>
function main() {
}
```

After this change, this will need to be split to multiple files.

PR #196, opened 2019-07-13
Use all strict options since HHVM4.8

lexidor:
Build failure: Difflib (already patched)

PR #199, opened 2019-07-24
Make use of the stricter options.

lexidor:


lexidor:
This does the same as #196, but without the rebase nonsense.

lexidor:
HHVM 4.17 ambiguous_lambda has been scaled back. `//UNSAFE` is nolonger supported anyway.

PR #201, opened 2019-08-01
Allow multiple schema versions

fredemmott:


jjergus:
> if the only differences are deletions

Does this happen often enough to warrant this extra complexity?

fredemmott:
Given we've not been updating the schema version appropriately, I don't know yet; I don't think this is overly complex, but if it happens to be worthless, happy to rip it out.

The alternative isn't great either:
- regen with current nightlies
- disable testing for any releases - only test on nightlies
- tag HHAST 4.17.0
- next monday/tuesday, re-enable release testing and remove the 'pre-release' flag from the 4.17.0 release

fredemmott:
Just a note from discussion: we can also do this in the opposite situation: if stuff is added - but not removed - we can codegen for the new version, but also allow the old version.

PR #209, opened 2019-08-29
update to AST version 2019-08-22 (used by 4.21+)

jjergus:
- removes `instanceof` and `__destruct`
- adds some "pocket universe" stuff

I'm planning to push this **after** the migration for `preg_match`, so I can make an HHAST release that has that migration and supports 4.15--4.20, and then another HHAST release that has the migration and supports 4.21.


src/__Private/codegen/CodegenRelations.hack
```diff
@@ -368,5 +368,7 @@ final class CodegenRelations extends CodegenBase {
     // https://github.com/hhvm/hhast/issues/151
     $_ = (this $x) ==> $x->flatten(dict[]) ?as this;
     $_ = (this $x) ==> $x->flatten(dict[]) as this;
+    /* HH_IGNORE_ERROR[4223] Only allowed in partial mode. */
+    (...) ==> null;
```
jjergus: https://github.com/facebook/hhvm/commit/ec4779427a1d786dd4ed820a2c82223465b92109 removed these from the codebase but the parser still supports it so I guess HHAST should too.

jjergus:
Travis passes on hhvm-nightly, others won't pass until next week.

PR #211, opened 2019-08-30
Consider inout and by-ref params to lambdas always used

### LINKED ISSUE #208: Unused local variable linter false positive
```HACK
$xxx = (inout vec<int> $p) ==> {
  $p[] = 1;
};
```
UnusedVariable: Variable is unusedHHAST(UnusedVariable) for $p.

ryangreenberg:
This fixes #208, where a variable inside a lambda expression that mutates an `inout` parameter of that expression is flagged as unused. This increases the potential for false negatives, an example of which is documented in the "clean" test cases.


jjergus:
Thanks for the quick fix!

PR #212, opened 2019-09-05
handle generics in resolve_type()

jjergus:
This changes the public API -- if we care about that, I can instead add this as a new function.

### DIFF
--- a/src/nodes/Script.hack
+++ b/src/nodes/Script.hack
@@ -86,4 +86,54 @@ final class Script extends ScriptGeneratedBase {
       },
     );
   }
+
+  /**
+   * Find all nodes that have generics and return both the nodes and their
+   * generics.
+   */
+  <<__Memoize>>
+  public function getNodesWithGenerics(): vec<(Node, TypeParameters)> {
+    $root = $this->getDeclarationsx();
+
+    // Sadly these don't have a consistent API so we need a separate Vec\map
+    // for each node type.
+    $nodes_with_maybe_generics = Vec\concat(
+      // classes/interfaces/traits
+      Vec\map(
+        $root->getChildrenOfType(ClassishDeclaration::class),
+        $n ==> tuple($n, $n->getTypeParameters()),
+      ),
+      // type declarations
+      Vec\map(
+        $root->getChildrenOfType(AliasDeclaration::class),
+        $n ==> tuple($n, $n->getGenericParameter()),
+      ),
+      Vec\map(
+        $root->getChildrenOfType(TypeConstDeclaration::class),
+        $n ==> tuple($n, $n->getTypeParameters() as ?TypeParameters),
+      ),
+      // functions/methods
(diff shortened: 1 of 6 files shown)


tests/ResolutionTest.hack
```diff
@@ -187,6 +187,8 @@ final class ResolutionTest extends TestCase {
         'Herp\\Derp',
         'Herp\\Derp',
       ),
+      // generics
+      tuple('<?hh namespace Foo; class Target<T> {}', 'T', 'T'),
```
jjergus: fails on master

src/nodes/Script.hack
```diff
@@ -86,4 +86,54 @@ final class Script extends ScriptGeneratedBase {
...
+  /**
+   * Find all nodes that have generics and return both the nodes and their
+   * generics.
+   */
+  <<__Memoize>>
+  public function getNodesWithGenerics(): vec<(Node, TypeParameters)> {
+    $root = $this->getDeclarationsx();
+
+    // Sadly these don't have a consistent API so we need a separate Vec\map
+    // for each node type.
```
fredemmott: Could be worth adding some - e.g. for functions/methods, IFunctionishDeclaration could be extended - MethodishDeclaration already has manual overrides with a generated base class, FunctionDeclaration would need the same

src/resolve_type.hack
```diff
@@ -12,12 +12,35 @@ namespace Facebook\HHAST;
 use namespace HH\Lib\C;
 use namespace Facebook\HHAST\__Private\Resolution;
 
-function resolve_type(string $type, Script $root, Node $node): string {
+enum ResolvedTypeKind: string {
+  GENERIC_PARAMETER = 'generic_parameter';
+  QUALIFIED_TYPE = 'qualified_type';
+}
+
```
fredemmott: Please split this to a separate file

jjergus: Hmm I'm OK either way but in general I don't really like putting everything in separate files -- it feels like taking a class and putting each method in a different file :/

src/nodes/Script.hack
jjergus: Yeah I tried that first but it turned out to be a lot of changes and this was much simpler. We can always refactor this later.

jjergus:
> This changes the public API -- if we care about that, I can instead add this as a new function.

Looks like [nothing is using this](https://github.com/search?q=org%3Ahhvm+%22resolve_type%22&type=Code) so we should be fine.

fredemmott:
It's also in `__Private`; while we probably want to start using this outside of HHAST, for as long as it's in __Private, breaking BC is fine

jjergus:
It's not -- only `resolve_name` (the implementation shared by `resolve_type` and `resolve_function` is __Private)

fredemmott:
oh whoops; thanks for catching that and checking for callers :)

PR #213, opened 2019-09-06
handle multiple bodyless namespaces

jjergus:
Right now HHAST's type resolution can't handle files with multiple namespaces without body, which is, unfortunately, legal syntax.

### DIFF
--- a/src/__Private/Resolution/get_current_namespace.hack
+++ b/src/__Private/Resolution/get_current_namespace.hack
@@ -10,23 +10,14 @@
 namespace Facebook\HHAST\__Private\Resolution;
 
 use type Facebook\HHAST\{Node, Script};
-use namespace HH\Lib\C;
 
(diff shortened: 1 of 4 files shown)


src/__Private/Resolution/get_current_namespace.hack
```diff
@@ -13,20 +13,12 @@ use type Facebook\HHAST\{Node, Script};
 use namespace HH\Lib\C;
 
 function get_current_namespace(Script $root, Node $node): ?string {
-  $namespaces = $root->getNamespaces();
-  if (C\is_empty($namespaces)) {
-    return null;
-  }
-  if ($namespaces[0]['statement']) {
-    return $namespaces[0]['decl']->getQualifiedNameAsString();
-  }
```
jjergus: I added the global namespace to the returned list so that we wouldn't need to handle these special cases everywhere (also this code wouldn't be correct for files with bodyless namespaces).

PR #214, opened 2019-09-06
fix missing autoimported types

jjergus:
also added a new ResolvedTypeKind for these, so callers can distinguish them if they want


src/resolve_type.hack
```diff
@@ -17,7 +17,57 @@ function resolve_type(
...
+    TypeStructure::class,
+    TypeStructureKind::class,
+    Vector::class,
+    WaitableWaitHandle::class,
+    XenonSample::class,
+    // not a legal type anymore but still supported by runtime
+    'callable',
+  ];
+
+  if (C\contains_key($autoimports, $type)) {
```
fredemmott: With the exception of ‘callable’, $autoimports should be fully qualified given they’re from ::class; sorry if I’ve missed something, but it looks like:

- this will only match on qualified names
- we shouldn’t be adding HH\ as its already qualified

Probably want a C\any( Str\ends_with

jjergus: You're right, I noticed it doesn't work when I tested more. I'll change them back to strings.

jjergus: The test coverage for various edge cases in hhast is not that great, but definition-finder has great coverage for these, so I end up finding most issues when I run definition-finder's tests on my changes in hhast :/

jjergus:
I guess callable is not really an autoimported type, so adding another special case for it :(

Let me know if you have a better idea, but it seemed to me like it wouldn't be correct to put it into any of the other "kinds".

PR #216, opened 2019-09-09
[PU] Starting linter for PU syntax

vsiles:
Disclaimer: CI will fail on this request since it requires Hack's experimental flag to work ok. I'm working on updating that so it is no longer needed, and I'll let you know when it's the case.

Creating linter to warn users if some of their files is using
reserved syntax for Pocket Universes.

Most of the syntax should just be erased from files right now.
It might happen that : and @ are close together, which will trigger
the linter, e.g. `function foo():@soft_annotation {...}`. In such cases,
the fix is to add a space in between

**Test is not working at the moment**, failing with this error:
```
~/hhast$ ./vendor/bin/hacktest tests/PocketAtomExpressionLinterTest.hack

Fatal error: Uncaught exception 'Facebook\HackTest\InvalidTestClassException' with message 'There must be exactly one test class in /home/ubuntu/hhast/tests/PocketAtomExpressionLinterTest.hack' in /home/ubuntu/hhast/vendor/hhvm/hacktest/src/Retriever/ClassRetriever.hack:72
Stack trace:
#0 /home/ubuntu/hhast/vendor/hhvm/hacktest/src/Runner/HackTestRunner.hack(42): Facebook\HackTest\ClassRetriever->getTestClassName()
#1 /home/ubuntu/hhast/vendor/hhvm/hsl/src/vec/transform.php(107): Closure$Facebook\HackTest\HackTestRunner::runAsync()
#2 /home/ubuntu/hhast/vendor/hhvm/hacktest/src/Runner/HackTestRunner.hack(42): HH\Lib\Vec\map()
#3 /home/ubuntu/hhast/vendor/hhvm/hacktest/src/HackTestCLI.hack(93): Facebook\HackTest\HackTestRunner::runAsync()
#4 /home/ubuntu/hhast/vendor/facebook/hh-clilib/src/CLIBase.hack(181): Facebook\HackTest\HackTestCLI->mainAsync()
#5 /home/ubuntu/hhast/vendor/hhvm/hacktest/bin/hacktest.hack(38): Facebook\CLILib\CLIBase::runAsync()
#6 /home/ubuntu/hhast/vendor/hhvm/hacktest/bin/hacktest(19): Facebook\HackTest\hack_test_main_async()
#7 (): Facebook\HackTest\hack_test_main_async_UNSAFE()
#8 (): Closure$__SystemLib\enter_async_entry_point()
#9 (): HH\Asio\join()
#10 (): __SystemLib\enter_async_entry_point()
#11 {main}
```

I'm not sure I understand why I get this error since I only have a single class in this file. Maybe it will ring a bell to more experienced developers :)

Also I added dummy code to `getCleanExamples` because the framework was complaining if the returned vector was empty. Maybe there's a better way to silence that since I don't need clean examples in here.

### DIFF
--- a/src/Linters/PocketUniversesLinter.hack
+++ b/src/Linters/PocketUniversesLinter.hack
@@ -1,0 +1,59 @@
+/*
+ *  Copyright (c) 2019-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+final class PocketLinterMessage {
+  const string Message =
+    'This is a reserved syntax for the Pocket Universes prototype.\n'.
+    'Only use it if you know  what you are doing.\n'.
+    'We advise you to remove it.';
+}
+
+// Displays warning for the :@atom_name case
+final class PocketAtomExpressionLinter extends ASTLinter {
+  const type TNode = PocketAtomExpression;
+  const type TContext = Script;
+
+  <<__Override>>
+  public function getLintErrorForNode(
+    Script $_context,
+    PocketAtomExpression $expr,
+  ): ?ASTLintError {
+    return new ASTLintError($this, PocketLinterMessage::Message, $expr);
+  }
+}
+
+// Displays warning for the Class:@Enum::foo case
+final class PocketIdentifierExpressionLinter extends ASTLinter {
+  const type TNode = PocketIdentifierExpression;
+  const type TContext = Script;
+
+  <<__Override>>
+  public function getLintErrorForNode(
+    Script $_context,
+    PocketIdentifierExpression $expr,
+  ): ?ASTLintError {
+    return new ASTLintError($this, PocketLinterMessage::Message, $expr);
+  }
+}
+
+// Displays warning for any Pocket Universes declaration/definition
+final class PocketEnumDeclarationLinter extends ASTLinter {
+  const type TNode = PocketEnumDeclaration;
+  const type TContext = Script;
+
+  <<__Override>>
+  public function getLintErrorForNode(
+    Script $_context,
+    PocketEnumDeclaration $expr,
+  ): ?ASTLintError {
+    return new ASTLintError($this, PocketLinterMessage::Message, $expr);
+  }
+}
--- a/src/Linters/PocketUniversesLinter.hack
+++ b/src/Linters/PocketUniversesLinter.hack
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2019-present, Facebook, Inc.
+ *  Copyright (c) 2017-present, Facebook, Inc.
  *  All rights reserved.
  *
  *  This source code is licensed under the MIT license found in the
--- a/src/__Private/LintRunConfig.hack
+++ b/src/__Private/LintRunConfig.hack
@@ -78,6 +78,9 @@ final class LintRunConfig {
     HHAST\GroupUseStatementsLinter::class,
     HHAST\GroupUseStatementAlphabetizationLinter::class,
     HHAST\NoWhitespaceAtEndOfLineLinter::class,
+    HHAST\PocketAtomExpressionLinter::class,
+    HHAST\PocketIdentifierExpressionLinter::class,
+    HHAST\PocketEnumDeclarationLinter::class,
   ];
 
   const vec<classname<BaseLinter>> NON_DEFAULT_LINTERS = vec[
--- a/tests/PocketAtomExpressionLinterTest.hack
+++ b/tests/PocketAtomExpressionLinterTest.hack
@@ -1,0 +1,22 @@
+/*
+ *  Copyright (c) 2019-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+final class PocketAtomExpressionLinterTest extends TestCase {
+    use LinterTestTrait;
+
+    protected function getLinter(string $file): PocketAtomExpressionLinter {
+        return PocketAtomExpressionLinter::fromPath($file);
+    }
+
+    public function getCleanExamples(): vec<(string)> {
+        return vec[tuple('<?hh if (true) {} else if (false) {}')];
+    }
+}
--- a/tests/PocketAtomExpressionLinterTest.hack
+++ b/tests/PocketAtomExpressionLinterTest.hack
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2019-present, Facebook, Inc.
+ *  Copyright (c) 2017-present, Facebook, Inc.
  *  All rights reserved.
  *
  *  This source code is licensed under the MIT license found in the
--- a/tests/PocketIdentifierExpressionLinterTest.hack
+++ b/tests/PocketIdentifierExpressionLinterTest.hack
@@ -1,0 +1,24 @@
+/*
+ *  Copyright (c) 2019-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+final class PocketIdentifierExpressionLinterTest extends TestCase {
+    use LinterTestTrait;
+
+    protected function getLinter(
+        string $file,
+    ): PocketIdentifierExpressionLinter {
+        return PocketIdentifierExpressionLinter::fromPath($file);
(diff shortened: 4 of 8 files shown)


jjergus:
> Test is not working at the moment, failing with this error:

This was because your autoload map was out of date (I'm guessing you renamed the class at some point?) -- I rebuilt it using `vendor/bin/hh-autoload` and now it no longer fails with this error. We have some logic in the autoloader that detects new classes without having to rebuild the autoload map, but I guess it doesn't detect newly removed classes so it ended up thinking there are multiple classes in that file. I'll make sure to document this somewhere.

Unfortunately it still fails even after fixing this :(

```
$ vendor/bin/hacktest tests/PocketAtomExpressionLinterTest.hack 
.E

1) Facebook\HHAST\PocketAtomExpressionLinterTest::testDirtyFixtures[0]
In file "/home/ubuntu/hhast/tests/examples/PocketAtomExpressionLinter/atom.hack.in": Failed to deserialize AST: IExpression expected, got Facebook\HHAST\ColonToken at /home/ubuntu/hhast/tests/examples/PocketAtomExpressionLinter/atom.hack.in:11:9
   10 function foo(): void {
   11     $x = :@bar;
     --------- ^ HERE

Summary: 2 test(s), 1 passed, 0 failed, 0 skipped, 1 error(s).
```

Looks like HHAST can't understand the PU syntax, even though the Hack parser/typechecker does. I'm not sure why that is, will investigate more tomorrow.

Random guess: HHAST generates its list of all valid syntax patterns by scanning tests in the HHVM repo. If some syntactic pattern doesn't occur in any tests, HHAST will not recognize it as valid. Maybe there are no HHVM tests for Pocket Universes? I'll check tomorrow.

vsiles:
> 
> 
> > Test is not working at the moment, failing with this error:
> 
> This was because your autoload map was out of date (I'm guessing you renamed the class at some point?) -- I rebuilt it using `vendor/bin/hh-autoload` and now it no longer fails with this error. We have some logic in the autoloader that detects new classes without having to rebuild the autoload map, but I guess it doesn't detect newly removed classes so it ended up thinking there are multiple classes in that file. I'll make sure to document this somewhere.

Indeed I renamed the file after I got a warning "single class in file, should be named ...." Nicely spotted !

> 
> Unfortunately it still fails even after fixing this :(
> 
> ```
> $ vendor/bin/hacktest tests/PocketAtomExpressionLinterTest.hack 
> .E
> 
> 1) Facebook\HHAST\PocketAtomExpressionLinterTest::testDirtyFixtures[0]
> In file "/home/ubuntu/hhast/tests/examples/PocketAtomExpressionLinter/atom.hack.in": Failed to deserialize AST: IExpression expected, got Facebook\HHAST\ColonToken at /home/ubuntu/hhast/tests/examples/PocketAtomExpressionLinter/atom.hack.in:11:9
>    10 function foo(): void {
>    11     $x = :@bar;
>      --------- ^ HERE
> 
> Summary: 2 test(s), 1 passed, 0 failed, 0 skipped, 1 error(s).
> ```
> 
> Looks like HHAST can't understand the PU syntax, even though the Hack parser/typechecker does. I'm not sure why that is, will investigate more tomorrow.
> 
> Random guess: HHAST generates its list of all valid syntax patterns by scanning tests in the HHVM repo. If some syntactic pattern doesn't occur in any tests, HHAST will not recognize it as valid. Maybe there are no HHVM tests for Pocket Universes? I'll check tomorrow.

There are a couple of examples in hphp/hack/test/pocket_universes but as I told you on the chat, their usage was behind heavy flagging, which should be removed pretty soon. The only remaining flag is a Hhbc one, see https://github.com/facebook/hhvm/blob/master/hphp/hack/src/hh_single_compile.ml#L363

vsiles:
I did a couple more tests and I realized that the experimental requirement is still present on the OSS side. I think we should wait for the next release to make sure all my patches are present (I still have a couple ones pending to review). Now that I know how to use the autoload, I'll ping you when the status of the pull request is passing. Thanks for all the feedback !

vsiles:
Did a small update today. Now I only have a single error test:
```
$ vendor/bin/hacktest tests/Pocke*
...E

1) Facebook\HHAST\PocketIdentifierExpressionLinterTest::testDirtyFixtures[0]
In file "/home/ubuntu/hhast/tests/examples/PocketIdentifierExpressionLinter/pu_ident.hack.in": Failed to deserialize AST: Expected nonnull, got null
   10 class C {
   11     enum E {
      ^ HERE

Summary: 4 test(s), 3 passed, 0 failed, 0 skipped, 1 error(s).
```

The other test files worked (`:@bar` is correctly spotted)  and `hh_single_compile` and `hhvm` supports correctly the PU syntax (you can run them with `hh_single_compile -v Hack.Lang.EnablePocketUniverses=1 foo.hack`). So I'm not sure why I get this error. On the test file, both `hh_single_compile` and `hhvm` are happy.

jjergus:
For some reason, the magic inside the HHAST codegen incorrectly inferred that each PU enum must have modifiers. I'm looking into it, but in the meantime you can unblock yourself by changing "enum" to "final enum".

fredemmott:
That probably means that all of the HHVM/Hack unit tests have modifiers; ideally tests should be added to cover other cases.

Just for HHAST, examples can be added to  src/__Private/codegen/data/*SyntaxExample.hack , and the codegen rebuilt (including relationships), but that's very slow.

jjergus:
> That probably means that all of the HHVM/Hack unit tests have modifiers; ideally tests should be added to cover other cases.

Nope :(

https://github.com/facebook/hhvm/blob/master/hphp/hack/test/pocket_universes/parsing/empty.good.php

jjergus:
I think we're hitting this case: https://github.com/hhvm/hhast/blob/master/src/__Private/codegen/CodegenSyntax.hack#L581

fredemmott:
Whoops, yep - that made sense back when we had `Missing` as a `Node`, but now we return `?T` instead of `T|Missing`

jjergus:
Oh and also `hphp/hack/test/pocket_universes` is not in the list of directories we look at, which is why we're hitting that code branch in the first place.

vsiles:
Travis CI is failing but I don't have anything wrong in local when I run `./vendor/bin/hacktest tests`. Should I wait until the AST is picking on the `hack/test/pocketuniverses` test suite  ?

jjergus:
Looks like it's failing because of a lint error. You can run `bin/hhast-lint` to reproduce.

One of the Travis runs also fails because it runs on an older HHVM version (4.21), you'll probably need to change `.travis.yml` so that it uses 4.23 instead of 4.21.

The issue we talked about (where our codegen incorrectly assumed all unknown types are non-nullable) is fixed in #222

vsiles:
Ok, then I'll rebase and try to update the .travis.yml tomorrow, and I'll let you know. Thanks for all the support !

fredemmott:
Please also update the HHVM version requirement in composer.json if you change the .travis.yml version - otherwise when the release is publishes, it will still be installable by users of older versions

jjergus:
> Please also update the HHVM version requirement in composer.json if you change the .travis.yml version - otherwise when the release is publishes, it will still be installable by users of older versions

Technically the only thing that depends on HHVM >= 4.23 is the test, does it make sense to bump the required HHVM version (and force people to upgrade or use old HHAST) because of that?

Maybe we could instead update the test to not run on older versions?

fredemmott:
> Technically the only thing that depends on HHVM >= 4.23 is the test, does it make sense to bump the required HHVM version (and force people to upgrade or use old HHAST) because of that?

I don't think we should support versions that we don't test on

> Maybe we could instead update the test to not run on older versions?

That would work - something like:

```Hack
if (\HHVM_VERSION_ID < 042300) {
  self::markTestSkipped("Test requires HHVM 4.23");
}
```

jjergus:
> I don't think we should support versions that we don't test on

Agreed in general, but in this case we're testing a Lint error on 4.23 which is a syntax error on < 4.23 so it seems pretty reasonable to skip the test like you suggested.

fredemmott:
Yeah, I"m fine with skipping a specific test that doesn't make sense - as long as the majority of the tests are run on the (oldest+newest) versions that composer permits

vsiles:
and now I'm lost :D I don't know enough your build system to know where the `markTestSkipped` should go :p

jjergus:
Oh OK, I think I can do that part. So just let me know when everything is ready, then I can merge this in and deal with the Travis/test issues.

vsiles:
Ok. I'll try to check the lint issue tomorrow and let you know .

vsiles:
@jjergus I just rebase on master + run the linter to make sure everyone's happy. Everything's ok on my side. If the current code is ok with you, fell free to merge and do the version magic you discussed above :)

jjergus:
Awesome, thanks!

jjergus:
https://github.com/hhvm/hhast/commit/aad064acd39f70819e0e9bb48ef61120ea5231bd adds the version check.

(There's still a test failure on the latest nightly builds, but that one is unrelated to PU.)

PR #217, opened 2019-09-10
add PHP override to .vscode/settings.json

jjergus:
This shouldn't be necessary because the Hack plugin is configured to recognize .php files that start with <?hh, but for some reason that doesn't work correctly when using remote VSCode.

Adding this override fixes it, and doesn't break anything elsewhere (even though it's redundant in most cases), so I think we can just add it here...?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,7 @@
 {
   "hack.hhastPath": "bin/hhast-lint",
-  "hack.useHhast": true
+  "hack.useHhast": true,
+  "files.associations": {
+    "*.php": "hack"
+  }
 }


fredemmott:
@PranayAgarwal have you used the remote editing features? Any idea why this is needed?

PranayAgarwal:
Hah I remember this config from years ago, when we had to add it to every project until this was fixed on the VS Code side. 

By remote I assume you mean VS Code's new remote functionality? I haven't had a chance to use it yet, but seems pretty weird that this is happening. It does work fine using the extension's remote features though (using SSH/Docker).

jjergus:
OK I'm gonna merge this in because it makes the remote experience much better, later we can try to figure out why it's needed and remove it after we fix the root cause.

jjergus:
@vsiles: In our "OnDemand AWS" system, right now this is added by the initialization script: https://github.com/hhvm/packaging/blob/master/aws/ondemand/hhvm/user.inc.sh#L22 (when you didn't see it, you probably just looked before the script finished)

But that pollutes `git status`, which is why I wanted to commit it here.

PR #220, opened 2019-09-13
RefToInoutMigration: implement optionalToRequired

jjergus:
This implements the missing logic for migrating things like `preg_replace_callback`, where an optional by-ref parameter became a non-optional inout parameter.

e.g. from:

```php
\preg_replace_callback(
  '/([a-z])/',
  fun('Str\\uppercase'),
  $str,
);
```

to:

```php
$__unused_inout = null;
\preg_replace_callback(
  '/([a-z])/',
  fun('Str\\uppercase'),
  $str,
  -1,
  inout $__unused_inout,
);
```

Unsurprisingly, most of the code here is to deal with whitespace and comments (see examples in the test).

I've put a lot of the code into reusable functions (`add_arguments`, `prepend_statement`, `whitespace_from_nodelist`), hopefully they could be useful for future migrations too.

### DIFF
--- a/src/Migrations/RefToInoutMigration.hack
+++ b/src/Migrations/RefToInoutMigration.hack
@@ -9,6 +9,8 @@
 
 namespace Facebook\HHAST;
 
+use namespace HH\Lib\{C, Vec};
+
 /**
  * Migrates builtin functions that originally accepted one or more parameters by
  * reference to the equivalent functions with inout parameters. For example,
@@ -23,6 +25,8 @@ final class RefToInoutMigration extends BaseMigration {
     'name' => Node,
     'args' => vec<IExpression>,
   );
+
+  const DUMMY_VARIABLE_NAME = '__unused_inout';
 
   <<__Memoize>>
   private static function getRules(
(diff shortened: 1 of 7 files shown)


src/Migrations/RefToInoutMigration.hack
```diff
@@ -129,11 +133,54 @@ final class RefToInoutMigration extends BaseMigration {
...
+
+    $new_call = add_arguments($n['root'], $n['call'], $args_to_add);
+
+    return $n['root']->replace($n['call'], $new_call)
+      |> prepend_statement($$, self::getDummyAssignment(), $new_call);
+  }
+
+  <<__Memoize>>
+  private static function getDummyAssignment(): IStatement {
+    return \HH\Asio\join(
```
fredemmott: image

src/Migrations/add_arguments.hack
```diff
@@ -0,0 +1,90 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
```
fredemmott: Perhaps this should be in __Private. Don't feel strongly

src/Migrations/prepend_statement.hack
```diff
@@ -0,0 +1,91 @@
...
+      ->withTrailing(whitespace_from_nodelist($root, $parent_block)['between']),
+  );
+
+  return $root->replace(
+    $parent_block,
+    NodeList::createMaybeEmptyList(
+      Vec\concat(
+        Vec\take($statements, $cur_idx),
+        vec[$new_statement],
+        Vec\slice($statements, $cur_idx),
```
fredemmott: `Vec\drop()`

src/Migrations/RefToInoutMigration.hack
jjergus: It's memoized so it happens at most once.

It wouldn't be too hard to convert all the functions to async, but honestly I'm scared of what edge cases could surface if the order of execution of everything is no longer deterministic :/

src/Migrations/prepend_statement.hack
jjergus: ummm...why do we have 2 HSL functions that do the same thing?

Also as far as I can tell Vec\drop is implemented less efficiently but that doesn't really matter here so I can change it.

src/Migrations/add_arguments.hack
jjergus: I put `whitespace_from_nodelist` in `__Private`, but `add_arguments` and `prepend_statement` seemed more generally useful so I put them here.

I'd argue that these functions have more right to be in a public namespace than all the migration/lint/etc. classes :)

src/Migrations/prepend_statement.hack
fredemmott: Well, slice() can also do the job of take(); probably made more sense when the HSL operated on Traversables rather than containers.

jjergus: To be fair, Vec\drop probably has more well-defined behavior for non-vec collections. Vec\slice (aka \array_slice) magically decides what to do with dict/keyset keys.

PR #221, opened 2019-09-13
add (limited) support for method calls

jjergus:
Only looks at the method name, so may result in false positives -- but that's OK for now (see comments inline).

If we need to implement more complicated transformations for methods later, we may need to fix that.

### DIFF
--- a/src/Migrations/RefToInoutMigration.hack
+++ b/src/Migrations/RefToInoutMigration.hack
@@ -9,7 +9,7 @@
 
 namespace Facebook\HHAST;
 
-use namespace HH\Lib\{C, Vec};
+use namespace HH\Lib\{C, Dict, Str, Vec};
 
 /**
  * Migrates builtin functions that originally accepted one or more parameters by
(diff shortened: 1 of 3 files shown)


src/Migrations/RefToInoutMigration.hack
```diff
@@ -158,14 +186,31 @@ final class RefToInoutMigration extends BaseMigration {
     return $n['root'];
   }
 
+  /**
+   * Same as above, but doesn't throw if the argument is not by-ref or inout.
```
fredemmott: nit: replace `above` with method name in case of re-ordering or new functions being added.

src/Migrations/RefToInoutMigration.hack
```diff
@@ -272,29 +317,43 @@ final class RefToInoutMigration extends BaseMigration {
...
+        // Note that we don't care about the type of the object, only about the
+        // method name. This can result in some false positives, but that's OK
+        // because & and inout can be used interchangeably right now.
+        $rule = self::getMethodRules()[$receiver->getName()->getCode()] ?? null;
+
+      } else {
+        continue;
+      }
 
       if ($rule is nonnull) {
```
fredemmott: with the `continue`, this should always be the case now?

jjergus: The `continue` is only for the "we don't know what this is" case (`$receiver` is neither of the 3 known types), this is still needed to handle the common case of a function/method for which there is no rule.

Issue #160, opened 2019-04-23
Add include/include_once/require linter


fredemmott:
Replace any of these with `require_once()`

PR #226, opened 2019-10-06
Write require_once linter, addresses issue #160

lexidor:
This linter is not yet done, but I want it to be known that I am writing it.

This is my very first attempt at writing a linter.
It does do the detection part well.
However, it fails on the rewriting part.
It consumes this whitespace:
```
    include     "string-like-expression";
^^^^       ^^^^^
```

I figured placing the whitespace between include and "string-like-expression" with a single space, should suffice.

However, this still kills all the indentation.

I hope that someone can drop a hint as a reply. :slightly_smiling_face:

### DIFF
--- a/src/Linters/PreferRequireOnceLinter.hack
+++ b/src/Linters/PreferRequireOnceLinter.hack
@@ -1,0 +1,48 @@
+/*
+ *  Copyright (c) 2019-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+final class PreferRequireOnceLinter extends AutoFixingASTLinter {
+  const type TContext = Script;
+  const type TNode = InclusionExpression;
+
+  <<__Override>>
+  protected function getTitleForFix(LintError $_): string {
+    return 'Prefer "require_once" over "include(_once)"" and "require"';
+  }
+
+  <<__Override>>
+  public function getLintErrorForNode(
+    Script $_context,
+    InclusionExpression $node,
+  ): ?ASTLintError {
+    if ($node is Require_onceToken) {
+      return null;
+    }
+
+    return new ASTLintError(
+      $this,
+      'Use require_once to require/include files',
+      $node,
+      () ==> $this->getFixedNode($node),
+    );
+  }
+
+  /**
+   * Could I get a helping hand here?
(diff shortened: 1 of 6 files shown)


lexidor:
Fixes #160

src/Linters/PreferRequireOnceLinter.hack
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Could I get a helping hand here?
+   * I want this to replace `include_once __DIR__ . '/../file.hack'` with `require_once __DIR__ . '/../file.hack'`.
+   * It consumes all the indentation.
+   * Is there something that feels like Node::getLeadingWhitespace()?
+   */
+  public function getFixedNode(InclusionExpression $node): ?Node {
+    return $node->withRequire(
+      new Require_onceToken(null, new NodeList(vec[new WhiteSpace(' ')])),
```
jjergus: Easiest way to preserve formatting is to copy the leading and trailing "trivia" from the original token.

- instead of `null` do `$node->getRequire()->getLeading()`
- instead of `new NodeList(...)` do `$node->getRequire()->getTrailing())`

This is also necessary to preserve comments. When you add a test case, also add some comments before/after/in the middle of the statement and verify they are preserved. With the existing code I suspect they wouldn't be.

lexidor:
Thank you, thank you, thank you. Will do as soon as I shouldn't be sleeping.

lexidor:
### Open to all advice

I have just finished writing the linter and the tests to go along with that.

This is my first attempt at doing something in hhast. Most of the things I did were heavily copy paste based. I would love to get to understand things like the testing system better.

Also, I was kind of paranoid about the example file needing to be executable. That is why I put in those `if (0 === 1)` around the bogus require statements and made the name of the trait super long. Will these files ever be executed? If no, do normal hack rules (like having unique names) apply? Or is each file / snippet looked at in isolation? Does the code in the file need to make sense or is any file that parses good enough? Do rules like no_toplevel_requires apply to these files?

lexidor:
### Git issues

Uhmmm, I see all the changes from the people I rebased on reflected in my change set. I think I screwed something up with rebasing. It looks like I needed to use `git push --force`. Not that vscode would have warned me about that in the first place. Is this branch still salvageable?

I know that the amount of actual work done is quite small, so it is not worth digging through git if that is an involved task. I have allowed changes by maintainers, so if this is an easy rookie mistake to fix, that would be an option. Else, I'll remake a PR without these issues using copy and paste.

jjergus:
> Is this branch still salvageable?

You can salvage any branch by deleting and recreating it from scratch, then `git push --force`. I did that here, please double check that nothing is missing.

lexidor:
@jjergus All files are there.

PR #228, opened 2019-10-07
refactor COMPATIBLE_SCHEMA_VERSIONS

jjergus:
+added compatibility between 2019-10-03-0001 and 2019-08-22-0001

Please sanity check that I didn't put in any of the compatibility rules backwards :)

### DIFF
--- a/src/__Private/is_compatible_schema_version.hack
+++ b/src/__Private/is_compatible_schema_version.hack
@@ -1,0 +1,45 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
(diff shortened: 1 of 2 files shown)


src/__Private/is_compatible_schema_version.hack
```diff
@@ -0,0 +1,45 @@
...
+    return true;
+  }
+
+  // Superset => subsets.
+  $compatible_versions = keyset[];
+
+  switch (SCHEMA_VERSION) {
+    case '2019-05-20-0001':
+      // No syntax was added, only removed, in the next schema version:
+      $compatible_versions[] = '2019-07-23-0001';
```
fredemmott: Are you expecting to backport this and some linter/migration changes? Otherwise we can probably just delete this branch

jjergus: Good point, yeah, this doesn't do anything in this branch...but maybe it has some documentation value + could be useful if we decide to backport something in the future...?

PR #230, opened 2019-10-11
add new compatible schema version

jjergus:


fredemmott:
Not quite compatible - travis found a problem

src/__Private/codegen/CodegenSyntax.hack
```diff
@@ -330,9 +330,17 @@ final class CodegenSyntax extends CodegenBase {
...
               $syntax['prefix'],
               $field['field_name'],
+              // Normally the JSON field always exists, even for nullable
+              // fields (it would just have "kind": "missing"), so we shouldn't
+              // need the following fallback. However, there is a special case:
+              // If the field is removed from the schema, then it will obviously
+              // not be present here. But we want to be able to treat the new
+              // schema as compatible with the previous version, which this
+              // fallback takes care of.
+              $spec['nullable'] ? " ?? dict['kind' => 'missing']" : '',
```
jjergus: ^ explanation here (also this is the only manual change, everything else is codegen)

jjergus: maybe worth putting `dict['kind' => 'missing']` into a const (`Node::MISSING` or something like that)?

PR #231, opened 2019-10-14
disallow_array_literal

lexidor:
I want to enable this setting in my project to discourage the use of plain php arrays.

This requires that all the dependencies also meet the `disallow_array_literal` requirement.

I replaced all `array(...)` and `[...]` expressions with their `darray`/`varray` counterpart.


jjergus:
Thank you! <3

I noticed that some of these functions are still declared to return `array`, we may need to change that to `varray`/`darray` too at some point (or even better, migrate straight to `vec`/`dict`).

PR #232, opened 2019-10-29
remove migration/linters for &$references

jjergus:
These are a parser error in HHVM 4.29+ so we can't support them, since
we get our AST from HHVM's parser.


composer.json
```diff
@@ -20,7 +20,7 @@
         "hhvm/hacktest": "^1.3"
     },
     "require": {
-        "hhvm": "^4.25",
+        "hhvm": "^4.29",
```
jjergus: Not strictly necessary, but I really want people with HHVM <4.29 to get the linters. I don't know if there's a better way to ensure that.

We can revert this to "^4.25" in a future release if

- anything useful (new linters/migrations) is added to HHAST so there's any value in doing so
- enough time has passed so that we're no longer concerned about people still using references

fredemmott:
travis.yml likely needs a change too

jjergus:
Passes on nightly, others should pass when 4.29 is released.

PR #233, opened 2019-10-31
Support `disable_lval_as_an_expression` AND `new_inference_lambda`

lexidor:


src/__Private/codegen/CodegenSyntax.hack
```diff
@@ -170,7 +170,7 @@ final class CodegenSyntax extends CodegenBase {
       );
     $intersected = C\reduce(
       $expanded,
-      ($acc, $i) ==> $acc = Keyset\intersect($acc, $i),
+      ($acc, $i) ==> Keyset\intersect($acc, $i),
```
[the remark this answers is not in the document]
fredemmott: Yep; also felt a little counterproductive - made me go looking for a missing `inout` :)

src/nodes/Token.hack
```diff
@@ -121,6 +121,7 @@ abstract class Token extends Node {
     $leading_list = __Private\fold_map(
       /* HH_FIXME[4110] use like-types when available*/ $json['leading'],
       ($j, $p) ==> Trivia::fromJSON($j, $file, $p, $source, 'Node'),
+      /* HH_FIXME[4297] use like-types when available*/
```
fredemmott: Like-types won't help with non-nested stuff - `as shape('width' => int, ...)` will work here.

src/Linters/GroupUseStatementsLinter.hack
```diff
@@ -270,6 +270,10 @@ final class GroupUseStatementsLinter extends AutoFixingASTLinter {
 
         if (
           Dict\filter($result[$kind][$namespace][1], ($n) ==> {
+            invariant(
+              $n is KeyedContainer<_, _>,
```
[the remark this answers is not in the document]
fredemmott: what is it if it's a string and what is it if it's a vec? Does sense to refactor to make the type of the variable consistent, instead of using `is`/`as`?

src/nodes/Token.hack
```diff
@@ -144,7 +144,8 @@ abstract class Token extends Node {
     $trailing_list = __Private\fold_map(
       /* HH_IGNORE_ERROR[4110] */ $json['trailing'],
       ($j, $p) ==> Trivia::fromJSON($j, $file, $p, $source, 'Node'),
-      ($j, $p) ==> $j['width'] + $p,
+      /* HH_FIXME[4297] use like-types when available*/
+      ($j, $p) ==> $j as shape('width' => int, ...) + $p,
```
fredemmott: is this fixme still needed?

lexidor: :slightly_smiling_face: no, I should have deleted it.

src/nodes/Token.hack
lexidor:
```
A linter threw an exception:
  Linter: Facebook\HHAST\AsyncFunctionAndMethodLinter
  File: /opt/lampp/htdocs/hhvm_repos/hhast/src/find_type_for_node.hack
  Exception: Facebook\HHAST\ASTDeserializationError
  Message: In file "/opt/lampp/htdocs/hhvm_repos/hhast/src/find_type_for_node.hack": Failed to deserialize AST: Expected shape('width' => int, ...), got HH\dict
```

This does not work unless the runtime setting for shape()'s are dict's is turned on.
The as fails at runtime.

src/nodes/Token.hack
```diff
@@ -144,7 +144,7 @@ abstract class Token extends Node {
     $trailing_list = __Private\fold_map(
       /* HH_IGNORE_ERROR[4110] */ $json['trailing'],
       ($j, $p) ==> Trivia::fromJSON($j, $file, $p, $source, 'Node'),
-      ($j, $p) ==> $j as shape('width' => int, ...) + $p,
+      ($j, $p) ==> $j as KeyedContainer<_, _>['width'] as int + $p,
```
fredemmott: ergh. would be really good for us to make dicts pass shape typechecks until we get a typed json_decode

src/nodes/Token.hack
lexidor:
```HACK
$j as KeyedContainer<_, _>['width'] as int
```

is almost the same thing as `$j as shape('width' => int, ...)['width']`, but it works with dicts too.

src/nodes/Token.hack
lexidor: I would be open for that. This only works because we are only indexing once. If we needed more from this structure, it would get pretty noisy pretty fast.

PR #235, opened 2019-11-02
Enable the NoPHPArrayLiteralsLinter as non-default

lexidor:


lexidor:
This may be a breaking change for many codebases. Therefore the linter is not in the default set. People who use `"builtinLinters": "all",` will get this change, but they have explicitly signed up for BC breaking linters.

PR #237, opened 2019-11-04
Implement NoFinalMethodInFinalClassLinter

### LINKED ISSUE #191: Add linter banning explicitly final methods if the class is already final
This can be a bit confusing in large files - if I see this, I assume that `foo()` can be overridden, but it can't if the class is final

```
public function foo(): void {
...
}

final public function bar(): void {
...
}
```

lexidor:
Fixes #191


lexidor:
This was a lot of fun.
As far as I understand it, this is done and meets the spec.

lexidor:
_Composer not found_ is the reason for test failure. :man_shrugging:
image

src/Linters/NoFinalMethodInFinalClassLinter.hack
```diff
@@ -0,0 +1,108 @@
...
+      $final->getTrailing()
+        |> $$->getCode() !== ' ' ? $$ : new NodeList(),
+    );
+    $first_trivia = NodeList::concat($first_token->getLeading(), $final_trivia);
+
+    return $context->replaceDescendant(
+      $first_token,
+      $first_token->withLeading($first_trivia),
+    );
+  }
```
jjergus: does this rely on the assumption that `final` is the first modifier? what does it do for something like `public /* foo */ final /* bar */ function...`?

src/Linters/NoFinalMethodInFinalClassLinter.hack
```diff
@@ -0,0 +1,108 @@
...
+      Str\format(
+        'Method %s is final in the class %s, which is also final. This is redundant.',
+        Str\trim($class->getName()->getCode()),
+        Str\trim($method->getFunctionDeclHeader()->getName()->getCode()),
+      ),
+      $method,
+      () ==> $this->getFixedNode($method),
+    );
+
+    echo $class->getName()->getCode().
```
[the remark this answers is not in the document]
lexidor: Yep :smile:

src/Linters/NoFinalMethodInFinalClassLinter.hack
```diff
@@ -0,0 +1,108 @@
...
+      ' is final and has the final method '.
+      $method->getFunctionDeclHeader()->getName()->getCode().
+      \PHP_EOL;
+  }
+
+  <<__Override>>
+  public function getTitleForFix(LintError $_): string {
+    return 'Remove final from methods in final classes';
```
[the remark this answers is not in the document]
lexidor: :+1:

src/Linters/NoFinalMethodInFinalClassLinter.hack
lexidor:
```HACK
final class FinalPublicVsPublicFinal {
  /*1*/final/*2*/ public/*3*/ function fpf(): void {}
  /*1*/public/*2*/ final/*3*/ function pff(): void {}
}
final class FinalPublicVsPublicFinal {
  /*1*//*2*/ public/*3*/ function fpf(): void {}
  /*1*//*3*/ public/*2*/ function pff(): void {}
}
```

The trail of final by prepending it to the trivia of the first token (or the second if the first token is `final`). Since I don't check /where/ final was in relation to the other token, this might end up changing the order of the trivia. If this is a problem, I will have to write logic that deals with appending it after the trailing travia of the token, to keep the relative order.

jjergus: Hmm maybe I'm missing something but it seems to me that your solution is unnecessarily complicated. Why not just take all the trivia and prepend it to the next token (or append to previous -- doesn't matter)? Something like:

```
$next = $ast->getNextToken($final_token);
return $node->replace($next, $next->withLeading(Vec\concat($final_token->getLeading(), $final_token->getTrailing(), $next->getLeading())));
```

plus some code to preserve the "single space" edge case, but that shouldn't make it much more complicated (probably just replace `$final_token->getTrailing()` with the same ternary expression you used)

jjergus:
Looks good, thanks!

PR #238, opened 2019-11-07
Strict mode only linter use plain hh

lexidor:
### Changes

 - Allow `<?hh` and `<?hh // strict` instead of just `<?hh // strict`.
 - Correct to `<?hh` when an autosuggest is requested.
 - Rename fix to `Use <?hh (default)` from `Use <?hh // strict`

I decided to still allow `// strict` for a while, since there is no good reason to break BC.
At some later point, we may decide to nolonger allow `// strict` and just keep hhvm's default.


lexidor:
@fredemmott Where is this json being generated from? https://github.com/hhvm/hhast/pull/238/commits/3bac6e7e9eabc3c15ea9114012f30f5fe5d9c3b4

You are the author of the expect file, but I don't know how it is getting these lint errors. It might very well be the you used a `<?hh` file to generate these errors (because that would trigger both a `NewlineAtEndOfFile` and a `StrictModeOnlyLinter`). Is this test still valid for you?

fredemmott:
LSPServerTest.hack creates an instance of the IDE integration, and a fake client; the JSON is LSP messages

The test is still valuable - it's checking that autofixes work for linters that aren't AST-based, like the whitespace linters.

PR #240, opened 2019-11-19
Remove an hhfixme

lexidor:


jjergus:
This one is intentional, we need an example of an untyped varargs lambda in the code, so that the HHAST codegen correctly infers that a varargs lambda can be untyped and generates a nullable getter.

We could maybe move it to a separate `//partial` file but I'm not sure if that would really be an improvement over a FIXME.

(feel free to reopen if you have other suggestions)

PR #242, opened 2019-11-30
Add new compatile schema version

lexidor:
This is the AST update.
[update AST* with reified information for PU](https://github.com/facebook/hhvm/commit/5193a75657ac268bad697510c6c7ce6e48eeeac0)
I don't know much about PU, but this seems to be a strict superset.


jjergus:
I think this is the wrong way around?

The new version adds syntax so it is not backwards compatible. But the previous version is compatible with the new one.

We'll also need to regenerate the codegen files I think.

fredemmott:
As long as this can still scan the HHIs and systemlib, it should be fine; PU's aren't a public feature yet, so the new syntax should not have a real impact

PR #243, opened 2019-12-02
fix AST version compatibility mapping

jjergus:
Also updating the comments to make it clear which version is the superset in each case.

If you prefer to keep backwards compatibility (ignoring the likely insignificant PU change), I can keep the technically-incorrect mapping (just clarify the comments) -- but that sounds to me like it is going to backfire sooner or later. I guess it could be reasonably safe if we remember to remove the incorrect mapping whenever the next non-trivial backward compatibility breaking change happens...?


src/__Private/is_compatible_schema_version.hack
```diff
@@ -29,21 +29,26 @@ function is_compatible_schema_version(string $other_version): bool {
...
+    case '2019-05-20-0001': // is a superset of:
+      // Succeeding versions:
       $compatible_versions[] = '2019-07-23-0001';
       break;
 
-    case '2019-10-03-0001':
-      // This schema is a strict superset of the previous version (nothing was
-      // removed, syntax for union and intersection typehints was added):
+    case '2019-10-03-0001': // is a superset of:
+      // Preceeding versions:
```
fredemmott: Preceding (though 'succeeding' is correct)

PR #247, opened 2019-12-12
faster hhast-inspect (chrome fix)

fredemmott:
hhast-inspect is borderline unusable in Chrome (though works fine in
Safari):
- it is incredibly slow to render, visibly rendering a line at a time
- hit tests take several seconds before it even touches the JS

Stop mirroring the AST in the DOM; instead, only put the tokens in the
DOM, but preserve the AST data in attributes. This gives us a flat and wide DOM tree instead of a super-deep one.


fredemmott:
@ssandler and @lexidor  are reporting this fixes the performance issues in Chrome and Opera

fredemmott:
No XHP yet because this code is namespaced :'(

PR #251, opened 2020-01-24
Make ShoutCaseEnumMembersLinter non-default

lexidor:
We would break BC otherwise.
Also, projects that use PascalCase shouldn't be forcefully convinced.


jjergus:
Thanks!

PR #253, opened 2020-01-31
Optimized PreferLambdasLinter for single-line expression.

### LINKED ISSUE #252: Optimize PreferLambdasLinter for single-line expression
Currently, single-line expression are auto fixed to `(...) ==> { return single-line-expression; }`
It will be great to improve PreferLambdaLinter to convert a single-line expression 
 to `() ==> single-line-expression` taking leverage of implicit returns.

cc: @ryangreenberg

joecheuk:
Resolves #252 

```
vendor/bin/hacktest tests/PreferLambdasLinterTest.hack 
....

Summary: 4 test(s), 4 passed, 0 failed, 0 skipped, 0 error(s).
```

### DIFF
--- a/src/Linters/PreferLambdasLinter.hack
+++ b/src/Linters/PreferLambdasLinter.hack
@@ -9,7 +9,7 @@
 
 namespace Facebook\HHAST;
 
-use namespace HH\Lib\C;
+use namespace HH\Lib\{C, Regex};
 
 final class PreferLambdasLinter extends AutoFixingASTLinter {
   const type TContext = Script;
@@ -71,8 +71,8 @@ final class PreferLambdasLinter extends AutoFixingASTLinter {
       null,
       new NodeList(vec[new WhiteSpace(' ')]),
     );
-    $body = $node->getBody();
-
+    $body = $this->simplifyBody($node->getBody());
+    //$body = $node->getBody();
     return new LambdaExpression(
       $attribute_spec,
       $async,
@@ -81,5 +81,25 @@ final class PreferLambdasLinter extends AutoFixingASTLinter {
       $arrow,
       $body,
     );
+  }
+
+  private function simplifyBody(ILambdaBody $body): ?ILambdaBody {
+    $content_excl_brackets = $body->getChildrenOfType(NodeList::class);
+    if (C\count($content_excl_brackets) !== 1) {
+      // Body has no content or more than one content groups, cannot be simplied.
+      return $body;
+    }
+    $statements = C\onlyx($content_excl_brackets)->getChildren();
+    if (C\count($statements) !== 1){
+      // Content has multiple statements, cannot be simplied.
+      return $body;
+    }
+    $statement = C\onlyx($statements);
+    if (!$statement is ReturnStatement) {
+      // Only a return statement can be simplified.
+      return $body;
+    }
+    // Extract the returned expression and return.
+    return $statement->getExpression();
   }
 }
--- a/src/Linters/PreferLambdasLinter.hack
+++ b/src/Linters/PreferLambdasLinter.hack
@@ -83,15 +83,10 @@ final class PreferLambdasLinter extends AutoFixingASTLinter {
     );
   }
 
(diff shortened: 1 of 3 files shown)


joecheuk:
Hi @lexidor , thanks for the quick response! Comments will be removed in the single line expression case, for example:

```
	$f5 = function(int $a): int { 
		// Test 123.
		return $a * $a; 
	};
**=>**
	$f5 = (int $a): int ==> $a * $a;
```

Do you have any other recommended approaches on handling comments in this case?

lexidor:
With my limited knowledge, I'd say:

Get the first token of the ReturnStatement using firstTokenx().
Get the leading travia of that using getLeadingTravia().
Then stitch that in front of the expression with some NodeList manipulation and withLeadingTravia().

There might be a prettier way, since I too am still learning about hhast, but this is how I've done things in the past.

lexidor:
The reason for this is `HH_FIXME` comments. If those get accidentally deleted, the code might nolonger pass the typechecker.

src/Linters/PreferLambdasLinter.hack
```diff
@@ -82,4 +82,24 @@ final class PreferLambdasLinter extends AutoFixingASTLinter {
       $body,
     );
   }
+
+  private function simplifyBody(ILambdaBody $body): ILambdaBody {
```
jjergus: If I'm reading everything correctly, `$body` here should be guaranteed to be `CompoundStatement` (because it's the result of calling `getBody()` on an `AnonymousFunction` object above). If you update the type annotation, that could simplify the code below (you can get the statements directly without the initial checks):

```hack
$statements = $body->getStatements()?->getChildren();
```

(or something like that)

jjergus:
Looking at the existing code in the linter, I think it's already throwing away some whitespace/comments (e.g. around the left parenthesis in `function(...`) so it probably doesn't make sense to try to make this one case 100% perfect.

So I guess, just do whatever seems like a reasonable tradeoff between correctness and complexity...

jjergus:
It might be possible to write a generic function that replaces any Node by one of its sub-nodes, while preserving all Trivia -- but that's outside the scope of this task. I might give it a try later.

src/Linters/PreferLambdasLinter.hack
joecheuk: Here is what I have after the update
```
  private function simplifyBody(CompoundStatement $body): ILambdaBody {
    $statements = $body->getStatements()?->getChildren();
    if (!$statements || C\count($statements) !== 1) {
      // Body has no statements, cannot be simplied.
      return $body;
    }
    $statement = C\onlyx($statements);
    if (!$statement is ReturnStatement) {
      // Only a return statement can be simplified.
      return $body;
    }
    return $statement->getExpressionx();
  }
```

Even though the code works, the return type isn't really as expected since LambdaExpression object is expecting an ILambdaBody as the body but we are returning an IExpression. I tried searching through existing classes that implements ILambdaBody but can't seem to find a good match to use, do we need to create a new class to handle that? (maybe a SingularStatement class vs CompountStatement)

joecheuk:
> With my limited knowledge, I'd say:
> 
> Get the first token of the ReturnStatement using firstTokenx().
> Get the leading travia of that using getLeadingTravia().
> Then stitch that in front of the expression with some NodeList manipulation and withLeadingTravia().
> 
> There might be a prettier way, since I too am still learning about hhast, but this is how I've done things in the past.

Since @jjergus is going to look into a better way of handling this later, I will leave the works out of the PR if that's ok with you.

src/Linters/PreferLambdasLinter.hack
jjergus: > LambdaExpression object is expecting an ILambdaBody as the body but we are returning an IExpression

Good catch, although it's weird why the same problem didn't affect the original version of the code.

I think you could just add `extends ILambdaBody` to the `IExpression` interface declaration -- as far as I can tell it's just an oversight that it isn't declared as such already.

joecheuk:
Updated, PTAL!

jjergus:
Looks good, thanks!

PR #258, opened 2020-02-10
fix DontAwaitInALoopLinter edge cases

### LINKED ISSUE #257: [ Bugreport ] False positive in DontAwaitInALoopLinter
The linter flags awaiting the expression to be iterated over as an await in a loop.

```HACK
<?hh

async function main_async(): Awaitable<void> {
    /*
    Don't use await in a loop
    Linter: Facebook\HHAST\DontAwaitInALoopLinter
    Location: /path/to/main.php:11:13
    Code:
    >    foreach (await awaitable_vec_of_int_async() as $int) {
    */
    foreach (await awaitable_vec_of_int_async() as $int) {
        $int;
    }
}

async function awaitable_vec_of_int_async(): Awaitable<vec<int>> {
    return vec[1];
}
```

Version information:
```
HipHop VM 4.43.0 (rel)
Compiler: 1580756284_957839074
Repo schema: 2cf430ca83644ef153159c3099262d5257773d80
HHAST: v4.33.5
```

jjergus:
- should not complain if await-ing in the part of the for/foreach loop that doesn't actually loop
- unless that loop is itself inside another loop
- the "closest function body" boundary wasn't correctly handled in the pretty-printer

Fixes #257

### DIFF
--- a/tests/examples/DontAwaitInALoopLinter/excluded.hack.expect
+++ b/tests/examples/DontAwaitInALoopLinter/excluded.hack.expect
@@ -1,0 +1,47 @@
+[
+    {
+        "blame": "await foo() ",
+        "blame_pretty": "Line 16:   foreach (await foo() as $_) {\nLine 17:     foreach (await foo() as $_) {",
+        "description": "Don't use await in a loop"
+    },
+    {
+        "blame": "      await foo() ",
+        "blame_pretty": "Line 16:   foreach (await foo() as $_) {\nLine 20:       await foo() as $_",
+        "description": "Don't use await in a loop"
+    },
+    {
+        "blame": "      await foo()",
+        "blame_pretty": "Line 16:   foreach (await foo() as $_) {\nLine 19:     foreach (\nLine 22:       await foo();",
+        "description": "Don't use await in a loop"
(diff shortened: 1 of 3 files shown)


tests/examples/DontAwaitInALoopLinter/excluded.hack.expect
```diff
@@ -0,0 +1,47 @@
+[
```
fredemmott: This is really hard to review... think it's time for a new expect format :) will do that separately

jjergus: In this case you just need to look at the line numbers, e.g. there is an error for line 16 + line 20, but there is no error for line 19 + line 20, as expected because the await on line 20 is inside the loop from line 16, but not inside the loop from line 19.

fredemmott: Yeah, it's doable, just not friendly - and not new with this PR, it's just I'm historically the one writing these so not used to being on this end of things :p

https://github.com/hhvm/hhast/issues/263 is one potential approach

PR #261, opened 2020-02-11
Correct outdated references to BSD+PATENTS license - we use MIT here.

### LINKED ISSUE #260: [ License ] Many non-code files refer to `the BSD-style license`
This might be intentional, but many `.expect` and `.in` files say that this library is actually BSD licensed, but the [LICENSE](https://github.com/hhvm/hhast/blob/master/LICENSE) files says MIT.

But then again, the [CONTRIBUTING.md](https://github.com/hhvm/hhast/blob/master/CONTRIBUTING.md#license) file says BSD again.

fredemmott:
fixes #260

### DIFF
--- a/tests/lsp/non-ast-in-line-code-action.json
+++ b/tests/lsp/non-ast-in-line-code-action.json
@@ -65,11 +65,11 @@
           {
             "range": {
               "start": {
-                "line": 11,
+                "line": 9,
                 "character": 17
               },
               "end": {
-                "line": 11,
+                "line": 9,
                 "character": 17
               }
             },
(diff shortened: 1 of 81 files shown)


fredemmott:
This change breaks tests/lsp/non-ast-in-line-code-action ... not sure why

fredemmott:
Ah, it asks for a fix for the issue at line <x> , but the line numbers are changed by the license header change

tests/lsp/non-ast-in-line-code-action.json
```diff
@@ -65,11 +65,11 @@
           {
             "range": {
               "start": {
-                "line": 11,
+                "line": 9,
                 "character": 17
               },
               "end": {
-                "line": 11,
+                "line": 9,
```
fredemmott: This is the only manual change - everything else is sed.

PR #264, opened 2020-02-12
Allow "." to be used as a root.

lexidor:
We have a project with ~50 hack files at the top level folder.
We want to lint them too, so we set our roots to `["."]`.
This makes it so hhast will lint all files in all folders.
We have added a override to not scan vendor.

This works in `vendor/bin/hhast-lint` already,
but the vscode intergration doesn't play nicely with this.
The paths the errors are reported on are not normalized.
Hhast reports errors at paths like
/path/to/./project/file.hack.
The problem explorer understands the non normal path.
However, the error go away when opening the file.
This commit strips the leading `./` that is added by having . in roots.

### DIFF
--- a/src/__Private/LintRunConfig.hack
+++ b/src/__Private/LintRunConfig.hack
@@ -155,9 +155,10 @@ final class LintRunConfig {
   public function getConfigForFile(string $file_path): self::TFileConfig {
     $roots = Vec\map(
       $this->configFile['roots'],
-      $s ==> Str\strip_suffix($s, '/').'/',
+      $s ==> Str\strip_suffix($s, '/').'/' |> Str\strip_prefix($$, './'),
     );
-    $file_path = Str\strip_prefix($file_path, $this->projectRoot.'/');
+    $file_path = Str\strip_prefix($file_path, $this->projectRoot.'/')
+      |> Str\strip_prefix($$, './');
     if (
       $roots !== vec[] &&
       !C\any($roots, $root ==> Str\starts_with($file_path, $root))


lexidor:
Our hhast-lint.json

```JSON
{
  "roots": [
    "."
  ],
  "builtinLinters": "all",
  "overrides": [
    {
      "patterns": ["*"],
      "disabledLinters": ["... our exceptions..."],
      "disableAllAutoFixes": false
    },
    {
      "patterns": ["./vendor/**"],
      "disableAllLinters": true
    }
  ]
}
```

lexidor:
The error would show in the problems pane
image
and clicking on the error takes you to the right file.
However, this squiggle wouldn't be here
image
and the error would be removed from the problem pane.

ssandler:
We also have `.` as a root because we have some files in the top level of the repo, and it gets awkward because `vendor/bin/hhast-lint top_level_file.php` does not actually run any linters on them, but `vendor/bin/hhast-lint ./top_level_file.php` does.

I just tried this change locally and it solves that problem too 👍

lexidor:
> I just tried this change locally and it solves that problem too :+1:

Good to know that I am not monkey patching OSS libs for just my own benefit. :100:

fredemmott:
is `.` interpreted as relative to the root or to the current directory?

lexidor:
If your repo lives at `/path/to/repo` putting "." in your roots will lint `/path/to/root/**`. You could write `some/folder/.` in roots, but that has the same effect as `some/folder/` or `some/folder`. This is only really useful for the dot as a catch all for linting files at the top level folder of the repo.

PR #267, opened 2020-02-18
Add `hhast-migrate --demangle-xhp-class-names`

fredemmott:
Replaces `-` with `_`, but not in attributes.

Attributes need to stay as-is, especially for `data-` and `aria-`
special-casing.

Also tested with:

```
fredemmott@fredemmott-mm1 xhp-lib % ~/code/hhast/bin/hhast-migrate --demangle-xhp-class-names src tests
fredemmott@fredemmott-mm1 xhp-lib % ~/code/hhvm-2/build/hphp/hack/bin/hh_client
No errors!
fredemmott@fredemmott-mm1 xhp-lib % tail -n 2 .hhconfig
disable_xhp_element_mangling=true
enable_xhp_class_modifer=true
```

Leaving `:` alone rather than doing a full demangle as it's likely
better for that to become a namespace separator than `__` in existing
code.

### DIFF
--- a/src/Migrations/DemangleXHP.hack
+++ b/src/Migrations/DemangleXHP.hack
@@ -1,0 +1,47 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+use namespace HH\Lib\Str;
+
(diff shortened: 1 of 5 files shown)


fredemmott:
cc @swahvay @ngavalas @johanoskarsson @tj09

fredemmott:
`$node->withFoo($bar)` preserves trivia of `$node`, but not any trivia on `Foo` - but if `$bar` is `$node->getFoo()->withBaz(...)` Foo's trivia is preserved.

fredemmott:
fc0c7b719dc7a716dee5b29a81691dc440a64e3f

PR #270, opened 2020-02-19
Add hhast-migrate --remove-xhp-child-declarations

fredemmott:
used to create hhvm/xhp-lib#224

Depends on and includes #268 - review last commit only

### DIFF
--- a/tests/examples/migrations/remove_xhp_child_declarations.hack.expect
+++ b/tests/examples/migrations/remove_xhp_child_declarations.hack.expect
@@ -1,0 +1,23 @@
+class :no:children {
+}
+
+class :no-replacement {
+}
+
+class :with-validation-trait{
+  use XHPChildValidation;
+
+  protected function getChildrenDeclaration(): XHPChild\Constraint {}
+}
+
+class :without-trait{
+
+  // This is /probably/ nonsense, but can't be sure
+  protected function getChildrenDeclaration(): XHPChild\Constraint {}
+}
+
+class :with-comment {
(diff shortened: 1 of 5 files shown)


tests/examples/migrations/remove_xhp_child_declarations.hack.expect
```diff
@@ -0,0 +1,23 @@
...
+}
+
+class :without-trait{
+
+  // This is /probably/ nonsense, but can't be sure
+  protected function getChildrenDeclaration(): XHPChild\Constraint {}
+}
+
+class :with-comment {
+  // Blah blah
```
jjergus: We should probably skip classes that don't have the `XHPChildDeclarationConsistencyValidation` trait, i.e. the migration would only apply to classes where we know for sure that the children declaration is redundant. Otherwise we might be removing children declarations that don't have a valid replacement.

fredemmott: I'd rather trust the user here - the usual flow should be:
- run `--add-xhp-child-declaration-methods` first
- test  thoroughly
- run `--add-xhp-child-declaration-methods` again, and check nothing's added
- run this on everything

> Otherwise we might be removing children declarations that don't have a valid replacement.

The foal is for children declarations to:
1. have no effect at runtime, i.e. be ignored by xhp-lib
2. be changed to parse errors

Removing unneeded is bad when we're at (2), but (1) needs  to happen first, and at that point it's a no-op

Issue #116, opened 2018-08-07
Add non-autofixing linter for `\HH\Asio\join()`


fredemmott:
Should:
 - handle if `join` is pulled in via a `use` statement
 - recommend refactoring to async functions and using `await` instead

lexidor:
Imma give this a shot.
I this post is prehistoric (:wave: hi, how's the future?) feel free to bump me or try it yourself.

lexidor:
This linter has been implemented in #274, closing.

PR #272, opened 2020-03-08
[ N/A ] Implement DontUseJoinLinter

lexidor:
I've chosen to shell out to `hh_client --find-refs \HH\Asio\join` because it is both easy to implement (for me) and very fast. `hh_client` already knows where all the functions are being referenced. It is just a matter of getting those references from its cache.

However, this linter would query your hh_client constantly if the files kept changing.

Let's disable this linter in the as-you-type-mode.


lexidor:
closes #116

lexidor:
The approach I chose is suboptimal.
If a change is made to n files (because of a previous linter), we will shell out to --find-refs n times. This becomes disastrous for performance. 

Also, if a file changes to have the same hash it had before, we won't shell out again. This can mean that the line numbers don't align and that ignored errors reappear. This can be fixed by keeping a dict from file to hash.

I'll have to think of an approach that does not rely on shelling out to --find-refs.

jjergus:
You could use `\Facebook\HHAST\resolve_function()` to avoid shelling out.

PR #274, opened 2020-03-14
Implement DontUseAsioJoinLinter

lexidor:
closes #272

### DIFF
--- a/src/qualified_name_is_fully_qualified.hack
+++ b/src/qualified_name_is_fully_qualified.hack
@@ -1,0 +1,20 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+/**
+ * @reviewer I'd like to put this function on QualifiedName.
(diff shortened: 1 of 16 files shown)


lexidor:
@jjergus this PR superceeds #272 and is ready for review.

lexidor:
closes #116

src/qualified_name_is_fully_qualified.hack
```diff
@@ -0,0 +1,14 @@
...
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+function qualified_name_is_fully_qualified(QualifiedName $name): bool {
```
jjergus: I think I remember from the original PR, you wanted to add this to the class?

You can do that by adding the class name to `CodegenSyntax::getKindsWithManualSubclasses()`, then `bin/update-codegen` will generate an abstract class that you can extend and add any methods you need.

lexidor: This dev ex improvement had to be postponed. See 6011173 for more details. It would conflict with another linter I wrote later.

lexidor: https://github.com/hhvm/hhast/pull/276#discussion_r393260011

jjergus:
Sounds good, thanks!

PR #276, opened 2020-03-14
Implement DataProviderTypesLinter

### LINKED ISSUE #234: Add linter against using PHP arrays etc as DataProvider return types
`vec` and single-element tuples are now supported; types like `array<array<string>>` can now be `vec<(string)>`

Additionally, I think (but need to test) that HackTest 2 will work properly with non-integer keys - i.e. you could do `dict<string, (string)>` to allow you to provide a custom informative key for error messages.

lexidor:
closes #234


lexidor:
Status update of what I've got so far.
```
Method Facebook\HHAST\LicenseHeaderLinterTest::getCleanExamples is a dataprovider for [testCleanExample] and has a return type of HH\vec<(HH\string)>
Method Facebook\HHAST\UseStatementWithLeadingBackslashLinterTest::getCleanExamples is a dataprovider for [testCleanExample] and has a return type of HH\vec<(HH\string)>
Method Facebook\HHAST\GroupUseStatementAlphabetizationLinterTest::getCleanExamples is a dataprovider for [testCleanExample] and has a return type of HH\vec<(HH\string)>
```
https://hastebin.com/raw/moqiwofipe

lexidor:
### Dependencies

Since I need to reflect on the DataProvider attribute, we now depend on HackTest being installed.
We could also hardcode `const string DATA_PROVIDER = 'Facebook\HackTest\DataProvider'` in the class and `HH_IGNORE_ERROR` the `$reflection_method->getAttributeClass(static::DATA_PROVIDER)`.
At runtime, this will be the same this, but it makes our code typecheck without having hacktest installed.
We do need a unit test that asserts that the dataprovider attribute still exists, in case hacktest changes the name.

lexidor:
closes #234

lexidor:
Status update of what I have got so far.

```
  Linter: Facebook\HHAST\UseCanonicalDataProvidersLinter
  Location: /path/to/hhast/tests/LSPServerTest.hack:72:2
  Code:
  >
  >  public function provideExampleExchanges(): array<varray<string>> 
Use a tuple as the value type of your return type
```

```
Use vec<_> or dict<_, _> return types
  Linter: Facebook\HHAST\UseCanonicalDataProvidersLinter
  Location: /path/to/hhast/tests/NodeAtPositionTest.hack:16:2
  Code:
  >  public function getExamples(
  >  ): array<(string, (int, int), classname<Node>, string)> 
```

lexidor:
Status update

I've finished the part of the linter that complains about your return type.
There are no auto fixes and no assignment checks between the data provider and the tests.

I do however feel like this linter is already providing value.
The question is "How many auto fixes do we want?".
If the data provider has got one return and it is an array literal, it could easily be transformed into a vec.
Tuplifying the elements if they are array literals with the same number of elements requires a ton of work.
Finding the best suited supertype for each element and such would easily balloon this linter to the size of a migration.

The assignment checks also rely on used traits and super classes, which would make this a cross file linter.
I think we can only do a best effort attempt and NOT go cross file.
If the provider can't be found in the test class, assume it is correct.
I'd like to implement this in a seperate PR. 

Opening this PR up for discussion.

lexidor:
@fredemmott What do you think?
How far should we go in writing autofixes?

src/Linters/UseCanonicalDataProviders.hack
```diff
@@ -0,0 +1,191 @@
...
+
+    $hhast_method = $hhast_methods[$method_name];
+
+    list($err, $_tuple_type) = $this->jugdeReturnTypes($hhast_method, $node);
+
+    if ($err is nonnull) {
+      return $err;
+    }
+
+    // TODO: Write validation for assignability from dataprovider to test method here.
```
fredemmott: An exact match, ignoring trivia would probably be appropriate - IMO it's fine to say that if the dataprovider gives a MySubclass, the test must take a MySubclass, not a MySuperclass.

src/qualified_name_is_fully_qualified.hack
```diff
@@ -0,0 +1,20 @@
...
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+/**
+ * @reviewer I'd like to put this function on QualifiedName.
+ *           Something like QualifiedName->isFullyQualified().
+ *           The file is a SignedSource file.
```
fredemmott: https://github.com/hhvm/hhast/blob/9f586c80055f73c17b95742ec97627c1d11a1deb/src/__Private/codegen/CodegenSyntax.hack#L690

Regen, then add manual subclass, e.g. https://github.com/hhvm/hhast/blob/master/src/nodes/ParameterDeclaration.hack

Make sure that you regen with the same version of HHVM/hh_parse that was last used for codegen - re-running the codegen should do nothing other than rename the base class and make it non-final

tests/UseCanonicalDataProvidersTest.hack
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public function getCleanExamples(): vec<(string)> {
+    return vec[
+      tuple(<<<'REQUIRE_ONCE_FILE'
+use type Facebook\HackTest\{DataProvider, HackTest};
+use function Facebook\FBExpect\expect;
+
+final class ATestClass extends HackTest {
+  public function provideAdditions(): vec<(int, int, int)> {
```
fredemmott: I'd like to see a test of `vec<(varray<int>)>` + darray

fredemmott: Also `dict<...., (a, b, c)>`

src/Linters/UseCanonicalDataProviders.hack
lexidor: Okay, just making sure, it is okay that this is a (false) positive / lint error:
Provider: `vec<(vec<string>, string)>`
Test: `function testThing(Traversable<string> $things, string $expected)`

tests/UseCanonicalDataProvidersTest.hack
lexidor: I'll add test cases for those.

lexidor: Fixed

src/Linters/UseCanonicalDataProviders.hack
fredemmott: IMO this is completely fine for a linter

fredemmott: one potential exception: perhaps the test function should be permitted to take generic parameters, in which case the types for that parameter are not checked.

src/qualified_name_is_fully_qualified.hack
lexidor: I'll have to postpone this QoL change, because DontUseAsioJoinLinter uses the same code...

src/Linters/UseCanonicalDataProviders.hack
```diff
@@ -0,0 +1,307 @@
...
+
+use namespace HH\Lib\{C, Dict, Math, Str, Vec};
+
+final class UseCanonicalDataProvidersLinter extends AutoFixingASTLinter {
+  const type TContext = ClassishDeclaration;
+  const type TNode = FunctionDeclarationHeader;
+  // dict<provider, vec<(test, provider again)>>
+  const type TProviders = dict<string, vec<(string, string)>>;
+
+  const string T_DATAPROVIDER = 'Facebook\\HackTest\\DataProvider';
```
fredemmott: Can you add a comment with a reason for this not being a `::class` constant? I'm guessing that it's so that it doesn't full-depend on hacktest?

src/Linters/UseCanonicalDataProviders.hack
```diff
@@ -0,0 +1,307 @@
...
+    this::TNode $node,
+  ): (?ASTLintError, ?TupleTypeSpecifier) {
+    $return_type = $hhast_method->getType();
+
+    // Only in partial files
+    if ($return_type is null) {
+      return tuple(
+        new ASTLintError(
+          $this,
+          'Add a typehint to your dataprovider',
```
fredemmott: `s/typehint/return type/`

Both more specific, and we don't really like the PHP 'typehint' term any more: it's not a hint in Hack

src/Linters/UseCanonicalDataProviders.hack
```diff
@@ -0,0 +1,307 @@
...
+      |> Vec\map($$, $li ==> $li->getItem());
+
+    $tuple_arity = C\count($tuple_types);
+    $test_arity = C\count($test_types);
+
+    if ($tuple_arity !== $test_arity) {
+      return new ASTLintError(
+        $this,
+        Str\format(
+          'Typeerror: Arity of test is %d, but the DataProvider has an arity of %d',
```
fredemmott: Type error

fredemmott: Also, perhaps "Test takes %d parameters, but DataProvider provides %d" ?

src/Linters/UseCanonicalDataProviders.hack
lexidor: > 
> ### Dependencies
> 
> Since I need to reflect on the DataProvider attribute, we now depend on HackTest being installed.
> We could also hardcode `const string DATA_PROVIDER = 'Facebook\HackTest\DataProvider'` in the class and `HH_IGNORE_ERROR` the `$reflection_method->getAttributeClass(static::DATA_PROVIDER)`.
> At runtime, this will be the same this, but it makes our code typecheck without having hacktest installed.
> We do need a unit test that asserts that the dataprovider attribute still exists, in case hacktest changes the name.

You are very much right. Let me add a comment about that.

lexidor:
I have not added this linter to the `default` or `all` levels for two reason.
 - It would break every build there is and this linter does not provide autofixes.
 - Scanning `src/` would a be huge waste of resources.

This may be reconsidered if we every decide to have a more granular way of opting in to newly developed linters.

lexidor:
https://github.com/hhvm/hacktest/pull/95 makes HackTest strict about the DataProvider attribute.
It previously consumed anything with name DataProvider, even from different namespaces.

src/Linters/UseCanonicalDataProviders.hack
```diff
@@ -0,0 +1,312 @@
...
+
+    $tuple_types = $data_provider_types->getTypes()->toVec()
+      |> Vec\map($$, $li ==> $li->getItem());
+
+    $tuple_arity = C\count($tuple_types);
+    $test_arity = C\count($test_types);
+
+    if ($tuple_arity !== $test_arity) {
+      return new ASTLintError(
+        $this,
```
fredemmott: Just to check:
- this raises an error on the dataprovider for test mismatch?
- IMO it should probably be on the test function instead, given you can use a single data provider for multiple...
- ... but doing that sucks and is slow in HHAST at the moment?

lexidor:
> Thanks, this looks good except for inline comments
> 
> Re performance of all/default, perhaps a dumb string-based approach could be used by overriding `shouldLintFile` ? e.g.
> 
>     * does filename contain 'test' (case-insensitive)?
> 
>     * does text contain 'HackTest'?

I think I will do a combination of the following. A test file is detected when the base name of the file contains 'test' and the file declares a class with the same name as the file.

This should cut down the cases where we scan no test classes.

lexidor:
The name of the linter is a misnomer at this point, since it does more correctness checks than just canonicalness. Open for suggestions for better names.

src/Linters/UseCanonicalDataProviders.hack
lexidor: The ASTLintError constructor takes a Node argument. Am I breaking the contract of getLintErrorForNode if I blame a different node than the one being given to me? If not, I can blame the FunctionDeclHeader of the test instead.

fredemmott: Hmm, yeah, it's currently a bit broken: getPrettyTextForNode takes a TNode but is called with others:

```
ASTLintError.hack
61:    return $this->linter->getPrettyTextForNode(/* HH_FIXME[4110] */$this->node);
```

The right thing to do is probably to make that take a Node, not a TNode, and make linters use `as ` etc.

The alternative fix is to make ASTLintError generic again, but that led to a lot of complicated spaghetti

lexidor: Let's stick with blaming the data provider for now.
We can make an issue about the getPrettyTextForNode() problem.
I am torn about the generic on LintError.
I'll have to weigh the up and downsides.

fredemmott: Yep, very much in favor of fixing that separately later, not blocking this :)

fredemmott:
DataProviderTypesLinter?

fredemmott:
Not merging, pending rename; I'm happy with TravisCI as long as we get signal on a reasonable subset of platforms. Travis have sadly removed/broken the 'retry' button :(

jjergus:
> Travis have sadly removed/broken the 'retry' button :(

It reappeared for me when I logged out and back in. Needs to be repeated when it disappears again :/

PR #281, opened 2020-04-09
[ Superceeded ] Activate the FinalOrAbstractClassLinter

lexidor:
This linter was not enabled with the `"enabledLinters": "all"` flag in the hhast-config.json file.


lexidor:
Let's see if git is smart enough to merge the file
We had two different hashes, but no conflicts in the body of the file
I updated the hash to the master one
If it fails, we'll know, because Travis checks the file sigs

PR #286, opened 2020-04-16
Make the version error correct using \HHVM_VERSION

lexidor:
before
AST version mismatch: expected '2020-04-06-0001' (40.53.0), but got '2020-04-14-0002

after
AST version mismatch: expected '2020-04-06-0001' (4.53.0), but got '2020-04-14-0002

### DIFF
--- a/codegen/version.hack
+++ b/codegen/version.hack
@@ -1,10 +1,12 @@
 /**
  * This file is generated. Do not modify it manually!
  *
- * @generated SignedSource<<ca9af36c1f9867e7412eebc1bc56ecab>>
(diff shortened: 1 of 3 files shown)


lexidor:
Tests will fail on nightlies because of this schema error

lexidor:
No this is wrong actually

The old code reffered to hhast's HHVM_VERSION_ID, not the root namespace one

lexidor:
We now use the value of `\HHVM_VERSION` => `Facebook\HHAST\HHVM_VERSION`
I left `Facebook\HHAST\HHVM_VERSION_ID` (no usages anymore), since external codebases might inspect this constant

PR #289, opened 2020-04-17
Remove MethodCallOnConstructorLinter

### LINKED ISSUE #288: Remove MethodCallOnConstructorLinter, since this is a parse error in modern Hack
```HACK
function main(): void {
    new AClass()->func();
}

class AClass {
    public function func(): void {}
}

```

image

lexidor:
fixes #288
closes #288


lexidor:
Removing the linter would break hhast-lint.json files that disable it explicitly.
The actually removal has to wait for the next breaking release.

PR #292, opened 2020-04-18
Remove dead code typehint nodes never start with `?T`

lexidor:
This ?T test never triggers, since `?` is its own token.
The means that `?Tk` explodes into `?` and `Tk`.
This caused a false positive while migrating hacktest.

```
DataProviderTypes: Potential type error: Parameter 1 on method testFilterNulls is of type
KeyedTraversable<Tk,?Tv>, but the DataProvider provides KeyedTraversable<mixed,mixed>
```

The rule is that you can assign anything to a generic, but `?Tv` destroys that.

### DIFF
--- a/src/Linters/DataProviderTypesLinter.hack
+++ b/src/Linters/DataProviderTypesLinter.hack
@@ -310,7 +310,6 @@ final class DataProviderTypesLinter extends AutoFixingASTLinter {
   }
 
   private static function isGeneric(string $string): bool {
-    return (Str\starts_with($string, 'T') || Str\starts_with($string, '?T')) &&
-      \ctype_alnum($string);
+    return (Str\starts_with($string, 'T')) && \ctype_alnum($string);
   }
 }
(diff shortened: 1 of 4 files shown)


jjergus:
> Add a failing test for ?Tk generics

Looks like the second commit accidentally removed the test? Can you add it back?

lexidor:
The test is a failing test.
It showed that the `?T` is not a typehint that accepts all types.
Adding it back would make the testsuite fail.

lexidor:
Such a shame that travis did not finish its run on that commit.
If it would have, you could have seen the :x:.

lexidor:
The test still fails now.
I just stripped out the `?T` check, since it was never true.

jjergus:
Oh, I see. But isn't it possible to keep the test but make it pass? i.e. change the expected output (expected errors + expected autofixes) to whatever we currently do, even if the current behavior is not optimal (there's still value in having a test, to document the current behavior and prevent someone from changing it unexpectedly)

jjergus:
Thanks! The test should also be useful in #294 where you can simply update it with the new behavior.

PR #293, opened 2020-04-18
Add File::isHackFile()

lexidor:


lexidor:
It appears like hh_parse _correctly_ confused php files with pure html for hack files with toplevel xhp.
I'll exclude these files from the autoloader paths, since it does not make sense to autoload them.

src/File.hack
```diff
@@ -9,11 +9,14 @@
...
 namespace Facebook\HHAST;
 
+use namespace HH\Lib\{C, Str};
+
 final class File {
   private function __construct(
     private string $path,
     private string $contents,
     private bool $isDirty,
+    private string $extension,
```
jjergus, line 19: Why not just compute the extension from the path inside `isHackFile()` (the only place where it's used), instead of passing it in? It would save a bit of code, and prevent people from accidentally introducing bugs if they pass in an inconsistent combination of path + extension somehow in the future.

lexidor: That is a good call indeed

PR #294, opened 2020-04-21
Write a special exception for `mixed` assign to `?`

lexidor:
https://github.com/hhvm/hsl/pull/132#discussion_r411768035

> I think the right fix here would be to add a special case to the linter so that it understands that mixed can match a nullable type (mixed is the only type that can, so this would be just that one special case).
> 
> The function is typed as generic `?T`, I don't think we want to change that to `mixed`.
>


lexidor:
This commit does however introduce this regression which I only just thought of just now.
`mixed` assign to `?int` is now true.
I should probably do something better here, since I believe that this is a terrible regression.

PR #295, opened 2020-04-22
sprintf() and invariant() now take ?arraykey for %s

lexidor:


lexidor:
Test can only succeed if https://github.com/hhvm/fbexpect/pull/28 is merged, however that test can only succeed if this is merged.

lexidor:
Superceeded by jjergus' fix

PR #296, opened 2020-04-26
Do not explode on PHP files

### LINKED ISSUE #290: [ Bug ] HHAST tries to lint `<?php` files
If HHAST sees a `.php` file it is instructed to lint, but this is a `<?php` file, this will almost always throw a fatal error, since `[]` array creation is illegal in Hack.

HHAST should skip `.php` files that do not contain a `<?hh` token.

lexidor:
closes #290


lexidor:
This change isn't very pretty, but it prevents hhast from going kaboom when it hits a `.php` file on accident that actually contains php.
I can lint in vscode-mode (tested by making a method in a final class final and seeing the yellow squigle) and using `bin/hhast-lint`, see travis output.

lexidor:
The :x: on FAILS-LINT is intentional.
It is the test case to show that `isHackFile()` is actually guarding the parser correctly.

I added `.vscode/**` to export-ignore, because I don't know if hhvm-autoload will try and parse the gunk php file when `composer require hhvm/hhast --dev`.

lexidor:
If I put it in a folder that hhvm-autoload can see, it will not be able to generate an autoloadmap.
It fails with a parse error.
I can place this anywhere, except for `src` and `tests`.
The directory needs to be added to `hhast-lint.json` to test effectively.

fredemmott:
can you add a `test-data/` top-level directory or similar?

hhast-lint.json
```diff
@@ -1,5 +1,5 @@
 {
-  "roots": [ "src/", "codegen/", "tests/" ],
+  "roots": [ "src/", "codegen/", "tests/", "test-data/" ],
```
fredemmott, line 2: Please remove this and add a test that invokes this and makes sure it's not linted

lexidor: Didn't see this earlier, resolved.

PR #298, opened 2020-06-09
Ban newline as first line of a block

### LINKED ISSUE #111: Add linter banning newline as the first line of a block
e.g.

```Hack
function foo() {

  bar();
```

lexidor:
fixes #111
closes #111


lexidor:
Blaming the next token over looks a bit weird, but I can not autofix a node I don't blame.
So echo is blamed for the newlines...

```HACK
function main(): void {

  echo "";
//^^^^
}

This doesn't hurt functionality, it is just surprising to see the yellow underline there, instead of below the `{` or actually on the whitespace.

**Comment by lexidor** [comment]
I sadly depend on Travis for testing after the rebase.
I am stuck with hhvm 4.56-LTS for a while.
If development becomes too difficult, I'll use vscode remote to a docker container with hhvm-latest.

**Comment by jjergus** [comment]
Would it make sense to exclude class/interface/trait declarations from this? I think most people prefer to have an empty line there (they could just not use the linter, but they might still want it for functions and for/if/... statement blocks).

Also any opinions about empty lines at the *end* of a block? Should the same linter handle that too?

**Comment by lexidor** [comment]
> Would it make sense to exclude class/interface/trait declarations from this?

That is a good call. I think it might be nice if this was configurable in some way. That way we can write linters that may have some overly agressive defaults to be tomed down, without needing to write the same linter twice (a strict and a lax version). Has storing linter specific settings in hhast-lint.json or `hhast-lint/MyLinterName.json` ever come up within Facebook?

**Comment by lexidor** [comment]
Will create a new branch + PR for this, because a lot of code has changed for newline reasons.

**Comment by lexidor** [comment]
> Would it make sense to exclude class/interface/trait declarations from this? I think most people prefer to have an empty line there (they could just not use the linter, but they might still want it for functions and for/if/... statement blocks).

With the config PR which landed a little while ago, we can now let people specify which block types should be ignored.

> Also any opinions about empty lines at the _end_ of a block? Should the same linter handle that too?

This would not be possible.
The linter targets LeftBraceToken and there is no super type for LeftBraceToken and RightBraceToken.
This could be done by targetting TokenWithFixedText, but 71 files contain `extends TokenWithFixedText`, so that would be a lot of needless function calls.

File: src/nodes/Script.hack [hunk 1]
```diff
@@ -125,8 +147,9 @@ final class Script extends ScriptGeneratedBase {
         if ($ns['name'] === '') {
           $ns['name'] = null;
         }
-        $inner =
-          __Private\Resolution\get_uses_directly_in_scope($ns['children']);
+        $inner = __Private\Resolution\get_uses_directly_in_scope(
```
lexidor, line 150: Accidental hackfmt

src/Linters/NoNewlinesAsTheFirstLineOfABlockLinter.hack
```diff
@@ -0,0 +1,120 @@
...
+ */
+
+namespace Facebook\HHAST;
+
+use namespace HH\Lib\{C, Str, Vec};
+
+final class NoNewlineAsTheFirstLineOfABlockLinter extends AutoFixingASTLinter {
+  <<__Reifiable>>
+  const type TConfig = shape(
+    'permit_newlines_after' => vec<string>,
```
lexidor, line 17: Bikeshed time, what do we name this key?

lexidor:
Will need to convert to draft.

- Support `do` token
- For loop support is broken, because of stopping on semi colons.

I might implement something in Script which can find the enclosing statement of a token. Don't know yet.

lexidor:
I'll create a new PR for this. Implementing this linter on the token level wasn't a smart idea.

PR #301, opened 2020-06-17
Add GitHub Actions CI

fredemmott:
- TravisCI is kind-of flakey now
- Adds MacOS testing
- Natively tests with ubuntu instead of docker


fredemmott:
Test run on my fork at https://github.com/fredemmott/hhast/actions/runs/138775681

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,63 @@
...
+            sudo add-apt-repository https://dl.hhvm.com/ubuntu
+            sudo apt-get install -y hhvm-nightly
+          elif [ "${{matrix.hhvm}}" = "latest" ]; then
+            sudo add-apt-repository https://dl.hhvm.com/ubuntu
+            sudo apt-get install -y hhvm
+          else
+            DISTRO=$(lsb_release --codename --short)
+            sudo add-apt-repository \
+              "deb https://dl.hhvm.com/ubuntu ${DISTRO}-${{matrix.hhvm}} main"
+            sudo add-apt-repository https://dl.hhvm.com/ubuntu
```
bigfootjon, line 38: Would it work if you switched the order of these `add-apt-repository` calls?

Then you could pull the "https://dl.hhvm.com/ubuntu" one out of the if statement completely

fredemmott: This is a bug; only the first one can be here.

This is currently invisible as this projects' minimum currently equals latest

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,63 @@
...
+        hhvm:
+          - 4.62
+          - latest
+          - nightly
+    runs-on: ${{matrix.os}}-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Composer
+        run: .github/workflows/install-composer.sh --install-dir=${{runner.temp}}
+      - name: Install HHVM (apt)
```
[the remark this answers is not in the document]
fredemmott: We also package for Debian; if someone wanted to specifically test there, they could in theory re-use this.

.github/workflows/build-and-test.yml
fredemmott: when 4.63 is released next week:
- the first line here would expose 4.62
- the second line would expose 4.63
- `apt-get install hhvm` would install 4.63

PR #302, opened 2020-07-02
Add compatible schema, skip PHP array literals linter on 4.65

fredemmott:
While I'm here, clean up old schema versions


fredemmott:
I left it in as I was thinking it could be valuable to keep recent ones there, but flush out the really old ones every so often

fredemmott:
eh, but it's dead code, and it's in source control. I'll do that.

PR #303, opened 2020-07-07
add github workflows from hsl-experimental

fredemmott:
Let's also get MacOS testing for HHAST

Ultimately once we're happy with these in a few components, the stuff
should be split out to reusable 'actions', rather than copying between
repos.


fredemmott:
This or something very much like it has already been merged

PR #304, opened 2020-07-15
Add unit test for new XHP class syntax

fredemmott:
This test in itself seems good, however I'm still getting parse
errors on the new xhp-lib. Need to investigate more, but this is
likely good regardless.


fredemmott:
xhp-lib fails to lint with out-of-source hhast as the class isn't in the autoload map, and starting up hh_client with default options leads to parse errors

Issue #313, opened 2020-08-04
Support custom settings per linter


lexidor:
Back to the original question, I'd add:

- per-linter configuration
  - we could just have EnumCaseLinter where you specify camel/shout/whatever as an option
  - indentation linter where you specify how many spaces/tabs
  - came up in https://github.com/hhvm/hhast/pull/298 ("ban newline as first line of a block" -- which kinds of blocks? classes/functions/statements)

_Originally posted by @jjergus in https://github.com/hhvm/hhast/issues/287#issuecomment-650308794_

lexidor:
Fixed by #314

PR #314, opened 2020-08-04
Support configs for individual linters

lexidor:


lexidor:
fixes #313

lexidor:
Funny that flipping the declarations makes the tests pass e441ff2.

src/Linters/BaseLinter.hack
```diff
@@ -9,11 +9,16 @@
...
 namespace Facebook\HHAST;
 
+use namespace Facebook\TypeAssert;
 use type Facebook\HHAST\File;
 use namespace HH\Lib\{C, Str};
 
 <<__ConsistentConstruct>>
 abstract class BaseLinter {
+  <<__Reifiable>>
+  const type TConfig as mixed = mixed;
```
jjergus, line 19: Maybe use `shape(...)` as the base type? Or are there any cases where we wouldn't want the config to be a shape?

src/Linters/BaseLinter.hack
```diff
@@ -23,6 +28,29 @@ abstract class BaseLinter {
...
+      throw new \InvalidOperationException(
+        Str\format(
+          '%s specified an unsupported config type. See previous exception:',
+          static::class,
+        ),
+        $e->getCode(),
+        $e,
+      );
+    }
+  }
```
jjergus, line 52: It would probably be better to be more strict here, instead of "anything that TypeAssert supports is valid config" (e.g. classes), it could be something like:

- config must be a shape (can be enforced statically, see above)
- valid shape field types: int, string, bool, maaaybe float
- vec or dict of valid shape type is also a valid shape type

This can't be enforced by TypeAssert but we could probably write a test for that (instead of the runtime check you have below [1]).

src/__Private/LintRunConfig.hack
```diff
@@ -240,6 +243,37 @@ final class LintRunConfig {
...
+    classname<TLinter> $classname,
+  ): ?TConfig where TConfig = TLinter::TConfig {
+    $config = idx(Shapes::idx($this->configFile, 'linterConfigs'), $classname);
+
+    if ($config is null) {
+      // Test for TypeAssert unsupported types, even when no config is supplied.
+      // This makes it impossible to accidentally ship a linter with a TypeAssert
+      // type that is not supported (if never tested with a config).
+      try {
+        $classname::typeAssertConfig(null);
```
jjergus, line 257: [1] This could be done in a test instead of at runtime (everytime any linter is invoked).

jjergus (approved):
Mostly looks good to me, added some comments inline.

fredemmott:
Bumping for previous comments, and a rebase

src/Linters/BaseLinter.hack
```diff
@@ -23,6 +28,29 @@ abstract class BaseLinter {
   public function __construct(private File $file) {
   }
 
+  public function setConfig(?this::TConfig $config): void {
+    $this->config = $config;
+  }
```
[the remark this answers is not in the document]
lexidor: Would it be required for `getConfig()` to be `final` to assure this immutability?
Just some random thought that popped into my mind when rereading my code.

lexidor:
The macos failure is the hh_server failing.
I don't know why the nightly builds have been stuck at running the tests for hours. (And why github hasn't killed the container yet.)

PR #315, opened 2020-08-04
Implement autofixing CRLF linter

### LINKED ISSUE #2: Add auto-fixing CRLF linter

lexidor:
fixes #2


jjergus:
Looks good to me, thanks!

PR #319, opened 2020-08-20
update schema (not backwards compatible), remove stuff

jjergus:
- First update that adds AST nodes after several removal-only updates, therefore this is only backwards compatible with a single previous AST version, which was only used in a few nightlies (no actual releases) -- so effectively, this update is not backwards compatible at all.
- Removes:
  - `NoPHPArrayLiteralsLinter`
  - `PHPArrayLiteralsMigration`
  - all 3 `PHPArrayTypehints` migrations
  - With some minor updates, we could keep supporting these (e.g. `array()` literals are still successfully parsed as "function calls", array typehints as regular generic types) -- if you think it would be useful, I can do that instead of removing them.
- Simplified some boilerplate code in `MigrationCLI`.
- All manual changes are in 2nd commit.


src/__Private/MigrationCLI.hack
```diff
@@ -387,6 +234,25 @@ class MigrationCLI extends CLIWithRequiredArguments {
     return $options;
   }
 
+  /**
```
fredemmott, line 237: Nice :)

jjergus:
> AIUI these need the old parser anyway to actually be able to read the array typehint, so using an old hhast + hhvm to do this migration seems reasonable

The parser can parse these just fine, it just thinks `array` is a class name (in typehints) or a function name (in literals). The only thing that can't be parsed is `[...]` literals which were already unsupported by the `--php-arrays` migration in recent HHAST releases.

So this update can be done without removing any features, I just wasn't sure it was worth doing vs just telling people to use specific HHAST versions.

PR #321, opened 2020-09-09
update schema (backwards compatible except not really)

jjergus:
The schema itself is backwards compatible, but the codegenned API changes aren't.

1. XHP children declarations still exist in the schema but HHVM no longer parses them (with default settings)
2. extra argument in `FunctionDeclarationHeader` constructor
3. changes to `MethodishTraitResolution` which we have documented as "whatever this is"

2 and 3 are more-or-less irrelevant but 1 is significant and probably warrants bumping the minimum HHVM version and making this a .y release.

I removed the XHP children migrations even though they have no Hack errors, because they no longer work (again, with default settings).

Alternatively, we could keep the XHP migrations if we revert https://github.com/hhvm/hhast/commit/77366473c594cdd07f7218d5569c944a589f29cc, set `src/codegen/data` as `export-ignore` and override the default setting in .hhconfig (which anyone trying to run the migration would also have to). Not sure if this is worth doing, but right now the range of HHAST versions that have all the XHP migrations available is quite small, so maybe...?


fredemmott:
This seems like time for a minimum version bump

tests/NodeTypesTest.hack
```diff
@@ -123,48 +123,8 @@ final class NodeTypesTest extends TestCase {
...
-    $lhs = $bin_expr->getOperand();
-    $lhs = expect($lhs)->toBeInstanceOf(XHPChildrenParenthesizedList::class);
-    expect($lhs->getCode())->toBeSame('(pcdata)');
-    $rhs = $bin_expr->getOperator();
-    expect($rhs)->toBeInstanceOf(StarToken::class);
-  }
-
   public async function testOldStyleXHPClassNodes(): Awaitable<void> {
-    $code = '<?hh class :foo { children (pcdata)*; }';
+    $code = '<?hh class :foo { }';
```
fredemmott, line 127: Add an attribute declaration here so that there's still some XHP-specific syntax inside the class

fredemmott (approved):

jjergus:
hmm now that I think of it, would it be better to bump the minimum HHVM version to ^4.73 (last week's release) instead of ^4.74?

The XHP migrations are already broken on HHVM 4.73 (with default settings), so we probably want anyone with HHVM 4.73 to get a HHAST release without the broken migrations.

jjergus:
Actually, I realized now that the `disable_xhp_children_declarations` setting only affects the typechecker, the runtime (hackc?) can still parse them -- so these migrations might actually still work fine, if people are willing to ignore Hack errors before they run the migration. I'll test it and if they indeed work fine, I'll revert the version bump and release this as .z

PR #324, opened 2020-09-23
use XHP for hhast-inspect

### LINKED ISSUE #323: HHAST-Inspect should use XHP
Currently doesn't as the code is namespaced, and XHP was unusable in namespaces until recently.

fredemmott:
This is mostly trivial, with a tiny bit of refactoring. Found a small
bug - pre/code open and close order were mismatched.

This should probably be split out to a separate repo; having HHAST
depend on xhp-lib isn't really acceptable, but neither is using HTML
strings when there's a better alternative.

fixes #323

### DIFF
--- a/src/__Private/Inspector/InspectorCLI.hack
+++ b/src/__Private/Inspector/InspectorCLI.hack
@@ -15,6 +15,20 @@ use type Facebook\CLILib\CLIWithRequiredArguments;
 use type HH\Lib\_Private\PHPWarningSuppressor;
 use namespace Facebook\CLILib\CLIOptions;
 
+use namespace /* HHAST_IGNORE_ERROR[UseStatementWithAs] */ Facebook\XHP\Core as x
+;
+use type Facebook\XHP\HTML\{
+  body,
+  code,
+  div,
+  head,
+  html,
+  pre,
+  script,
+  span,
+  style,
+};
+
 final class InspectorCLI extends CLIWithRequiredArguments {
   private ?string $outputPath = null;
   private bool $open = false;
@@ -71,7 +85,7 @@ final class InspectorCLI extends CLIWithRequiredArguments {
 
     \file_put_contents(
       $output,
-      self::getHTMLHeader().$this->getHTMLForNode($ast).self::getHTMLFooter(),
+      await $this->getXHPDocument($ast)->toStringAsync(),
     );
 
     print $output."\n";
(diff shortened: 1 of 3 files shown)


src/__Private/Inspector/InspectorCLI.hack
```diff
@@ -15,6 +15,20 @@ use type Facebook\CLILib\CLIWithRequiredArguments;
 use type HH\Lib\_Private\PHPWarningSuppressor;
 use namespace Facebook\CLILib\CLIOptions;
 
+use namespace /* HHAST_IGNORE_ERROR[UseStatementWithAs] */ Facebook\XHP\Core as x
+;
```
fredemmott, line 19: Hackfmt bug I guess :)

src/__Private/Inspector/InspectorCLI.hack
```diff
@@ -15,6 +15,20 @@ use type Facebook\CLILib\CLIWithRequiredArguments;
 use type HH\Lib\_Private\PHPWarningSuppressor;
 use namespace Facebook\CLILib\CLIOptions;
 
+use namespace /* HHAST_IGNORE_ERROR[UseStatementWithAs] */ Facebook\XHP\Core as x
```
fredemmott, line 18: @lexidor previously suggested a whitelist for UseStatementWithAs - this should be configurable, but we potentially want to whitelist Facebook\XHP by default

jjergus (approved):

lexidor (approved):
Code changes are indeed mostly trivial.
How about that circular dependency?
Should the inspect functionality be its own _small_ package?

lexidor:
Test failure is strange.
Testing on ubuntu against 4.72 resulted in the following message:
```
Get:30 https://dl.hhvm.com/ubuntu bionic-4.72/main amd64 Packages [1272 B]
Fetched 20.6 kB in 3s (8004 B/s)
+ sudo apt-get install -y hhvm
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
hhvm is already the newest version (4.75.0-1~bionic).
0 upgraded, 0 newly installed, 0 to remove and 27 not upgraded.
```

Mac fails (correctly), since xhp 4 can't run on hhvm 4.72.

fredemmott:
Closing: hhast-inspect is being split to a new repository for this because of the circular dependency. Basically ready to go, waiting for final approval on the new repo

PR #326, opened 2020-10-02
Package private linter to enforce module boundaries

YashiKamboj:
The linter is enforcing the following rules:
1. If a file has no top level namespace, then it cannot use any `__Private` or `_Private` artifacts.
2. If a file has a top level namespace, then it can only use `__Private` or `_Private` artifacts from within the top level parent namespace.

Would be happy to add more cases if I am missing something here to enforce namespace boundaries. Thank you!


composer.json
```diff
@@ -25,7 +25,7 @@
         "hhvm/hsl": "^4.25",
         "hhvm/hsl-experimental": "^4.58.0rc1",
         "hhvm/hsl-io": "^0.2.0",
-        "hhvm/hhvm-autoload": "^2.0.4|^3.0",
```
YashiKamboj, line 28: Tried running tests with this setting and I could not get it working, hhvm-autoload was 3.1 for HHVM 4.72. Not sure if this change is required, just wanted to add it for review.

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+  NamespaceDeclaration,
+  NamespaceGroupUseDeclaration,
+  NamespaceToken,
+  NamespaceUseClause,
+  NamespaceUseDeclaration,
+  Node,
+  QualifiedName,
+  Script,
+  TypeToken,
+};
```
fredemmott, line 27: Heh, we should add a linter to pick these all up - as you're in the HHAST namespace, these are all usable without an explicit `use type`

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+use namespace HH\Lib\{C, Str, Vec};
+
+type use_definition_parts_t = shape(
```
fredemmott, line 30: - `type TUseDefinitionParts` would be following the convention here
- these should probably be type constants on the class instead of top-level types in the HHAST namespace

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+  public function getLintErrorForNode(
+    this::TContext $context,
+    this::TNode $node,
+  ): ?ASTLintError {
+    // Get the current namespace for the file.
+    $current_namespace_for_file = vec[];
+    $top_level_namespace = $this->getTopLevelNamespace($node);
+    if ($top_level_namespace is nonnull) {
+      $current_namespace_for_file = Str\split($top_level_namespace, '\\');
+    }
```
fredemmott, line 55: as you're getting this out of a list of nodes, it could be better to directly return a list of parts, instead of flattening a list to a string then back again

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+
+    // Process all use statements present in the file.
+    $use_definition_context_shape = $this->getUseDefinitions($context);
```
fredemmott, line 58: If you need something low level, `Facebook\HHAST\__Private\get_uses_directly_in_scope` - but I think `$script->getNamespaces()` gives you everything you need here without needing that; it can also reimplement getToplevelNamespace

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+    $use_definition_context = Shapes::toArray(
+      $use_definition_context_shape,
+    );
+
+    // Go through use definition to check for potential violations of package private.
+    foreach ($use_definition_context as $_key => $use_definition_context_value) {
```
fredemmott, line 64: nit: I'd skip the `$_key =>` entirely - if it's unused and you don't need to specify it, don't, especially when the nameisn't descriptive

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+      foreach ($use_definition_context_value as $_inner_key => $use_definition_context_inner_value) {
+        if (!$this->isPrivateNamespacePathAllowed(
+          $use_definition_context_inner_value['parts'],
+          $current_namespace_for_file
+        )) {
+          return new ASTLintError(
+            $this,
+            'Package private artifacts are private to the parent package',
```
fredemmott, line 72: Hack doesn't currently have the concept of packages, though there are proposals to change that. This linter should consistently refer to namespaces, not packages.

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+            $use_definition_context_inner_value['node'],
+          );
+        }
+      }
+    }
+
+    foreach ($node->getDescendantsOfType(QualifiedName::class) as $qualified_name_token) {
+      if ($qualified_name_token->hasParts()) {
+        $qualified_name_parts = $this->getQualifiedNameFlatten(
```
fredemmott, line 81: `Facebook\HHAST\resolve_function()` and `Facebook\HHAST\resolve_type()`, don't reimplement this

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+      }
+    }
+    return true;
+  }
+
+  /**
+   * This method returns the namespace for the file defined at the very top.
+   * eg: namespace TestNamespace; returns TestNamespace.
+   */
+  private function getTopLevelNamespace(this::TNode $node): ?string {
```
fredemmott, line 137: mentioned above: think this is unneeded, you just need `$script->getNamespaces()`

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+    $namespace = null;
+    foreach ($node->getDescendantsOfType(NamespaceDeclaration::class) as $namespace_token) {
+      $namespace = $namespace_token->getName();
+      if ($namespace is null) {
+        continue;
+      }
+      break;
+    }
+    return $namespace?->getCode();
```
fredemmott, line 146: if there's a reason you're reimplementing this, you must build it up from the tokens and use getText, or whitespace/comments can mess this up.

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  /**
+   * In some cases we don't have a fully qualified name.
+   * For example TestNamespace::testMethod() does not have a fully qualified name.
+   *
+   * To resolve these we extract the `use type`, `use function`, `use namespace` definitions from the file.
+   */
+  private function getUseDefinitions(Script $script): package_private_use_definitions_t {
```
fredemmott, line 155: also in $script->getNamespace(), uneneded

src/Linters/PackagePrivateLinter.hack
```diff
@@ -0,0 +1,318 @@
...
+        }
+      }
+
+      if (!$absolute_path && C\count($parts) > 0) {
+        /**
+         * get the absolute path by checking the following:
+         * 1. Is the path mentioned in qualified name relative to one of the namespace in use namespace statement(s).
+         * 2. Is the path mentioned in qualified name relative to the namespace of the current file.
+         */
+        $path_found = false;
```
fredemmott, line 296: this isn't complete, especialyl if there are multiple namespace blocks - the resolve_type/resolve_function functions mentioned above do this

src/Linters/PackagePrivateLinter.hack
fredemmott: This also applies to the linter name

src/__Private/LintRunConfig.hack
```diff
@@ -82,6 +82,7 @@ final class LintRunConfig {
     HHAST\PocketEnumDeclarationLinter::class,
     HHAST\PreferRequireOnceLinter::class,
     HHAST\ConsistentLineEndingsLinter::class,
+    HHAST\PackagePrivateLinter::class,
```
fredemmott, line 85: cc @jjergus @lexidor - IMO this one should go in NON_DEFAULT_LINTERS

tests/PackagePrivateLinterTest.hack
```diff
@@ -0,0 +1,24 @@
...
+final class PackagePrivateLinterTest extends TestCase {
+  use LinterTestTrait;
+
+  protected function getLinter(string $file): BaseLinter {
+    return PackagePrivateLinter::fromPath($file);
+  }
+
+  public function getCleanExamples(): vec<(string)> {
+    return vec[
+      tuple('<?hh function test();'),
```
fredemmott, line 21: add tests in namespaces using private definitions

fredemmott (changes requested):
How common are deeply nested namespaces? I'm wondering if defaulting to two levels for the common `namespace Vendor\Library;` would be reasonable

For example, `HH\Lib\IO` depends on `HH\Lib\_Private\_IO`; this would be banned by this linter, but permitted if the default was to scope to `Vendor\Library`

---

Add tests for both _Private and __Private - both should be treated similarlyly here

fredemmott:
Thanks - this is good overall, but doing a lot of work that's already done elsewhere - hopefully it can be directly re-used or extended if needed :)

src/Linters/PackagePrivateLinter.hack
YashiKamboj: Wrote the linter in Slack workspace earlier and forgot to remove these here. Thanks for pointing this!

src/Linters/PackagePrivateLinter.hack
YashiKamboj: makes sense, with your other suggestion of using `getNamespaces()`, we have completely removed this `getTopLevelNamespace()` function.

src/Linters/PackagePrivateLinter.hack
YashiKamboj: This was a game changer, thank you!

src/Linters/PackagePrivateLinter.hack
YashiKamboj: New to Hack, this makes sense.

src/Linters/PackagePrivateLinter.hack
YashiKamboj: YASS!
Have an issue resolving scope expressions and name expressions, wrote my own logic for that. Let me know if that sounds right.

src/Linters/PackagePrivateLinter.hack
YashiKamboj: removed this function entirely.

src/Linters/PackagePrivateLinter.hack
YashiKamboj: removed this function, and considered multiple namespaces in a single file.

tests/PackagePrivateLinterTest.hack
YashiKamboj: Assuming you meant rewriting the tuple part here to include `private` namespace.

YashiKamboj (commented):
> How common are deeply nested namespaces? I'm wondering if defaulting to two levels for the common `namespace Vendor\Library;` would be reasonable
> 
> For example, `HH\Lib\IO` depends on `HH\Lib\_Private\_IO`; this would be banned by this linter, but permitted if the default was to scope to `Vendor\Library`
> 
> Add tests for both _Private and __Private - both should be treated similarlyly here

Not very common IMO, seen mostly two level nested here at Slack. In your example:
>`HH\Lib\IO` depends on `HH\Lib\_Private\_IO`; this would be banned by this linter -> With the current logic , this **won't** be banned by the linter because we currently check the root(first parent) directory to be the same which is `HH` in this case and hence they belong to the same parent namespace.

YashiKamboj:
> Thanks - this is good overall, but doing a lot of work that's already done elsewhere - hopefully it can be directly re-used or extended if needed :)

Thanks for the review Fred! I am mostly done making the suggested changes, and should have a review out tomorrow.

tests/NamespacePrivateLinterTest.hack
```diff
@@ -0,0 +1,24 @@
...
+final class NamespacePrivateLinterTest extends TestCase {
+  use LinterTestTrait;
+
+  protected function getLinter(string $file): BaseLinter {
+    return NamespacePrivateLinter::fromPath($file);
+  }
+
+  public function getCleanExamples(): vec<(string)> {
+    return vec[
+      tuple('<?hh function test();'),
```
YashiKamboj, line 21: Will be fixing this.

src/Linters/NamespacePrivateLinter.hack
```diff
@@ -0,0 +1,148 @@
...
+
+					$parent_node = $context->getParentOfDescendant($qualified_name_token);
+					if ($parent_node is ScopeResolutionExpression || $parent_node is NameExpression) {
+						$fully_qualified_name_for_current_token = $this->resolve_scope(
+							$name_token_key,
+							$ns['uses']['namespaces'],
+							$current_namespace
+						);
+					} else if ($parent_node is FunctionCallExpression) {
+						$fully_qualified_name_for_current_token = resolve_function(
```
YashiKamboj, line 72: `resolve_function()` and `resolve_type()` also use `get_uses_directly_in_scope()` under the hood, same as used by `getNamespaces()` - not sure if there is an efficient way of doing this using the existing code.

src/__Private/LintRunConfig.hack
YashiKamboj: Curious: How is a non default linter activated?

lexidor: The `builtinLinters` options in the hhast config has three possible values. `all`, `default`, and `none`. When set to `all`, the non defaults are active. When set to `default`, the non defaults are excluded. `none` doesn't activate any linters. All modes allow `extraLinters` to be set as well.

tests/PackagePrivateLinterTest.hack
fredemmott: adding another tuple to the same vec :)

src/Linters/NamespacePrivateLinter.hack
fredemmott: It's probably worth us moving that to scripts/nodes and caching it... but that'll eat more memory :(

composer.json
fredemmott: These changes should be unneeded, assuming you're letting composer install them - these constraints should always the latest version. If it's using stale versions, try deleting vendor/ and composer.lock, and trying again

src/Linters/NamespacePrivateLinter.hack
```diff
@@ -0,0 +1,148 @@
...
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+use namespace HH\Lib\{C, Str, Vec};
+final class NamespacePrivateLinter extends ASTLinter {
+	const type TContext = Script;
```
fredemmott, line 13: run hackfmt with default settings - should end up being a 2-space indent

tests/examples/NamespacePrivateLinter/namespace_private_in_no_ns.php.expect
```diff
@@ -0,0 +1 @@
+[]
```
fredemmott, line 1: if they're short enough, tests with no errors should generally be tuples in the `getCleanExamples()` part of the test

fredemmott (approved):
Assuming tests are clean when rebased, it'd be good to fix these nits - but I'll merge then fix them up if they're all that's left

fredemmott:
Probably replaced by #327

PR #327, opened 2020-10-23
Package private linter

fredemmott:
Replaces #326

Creating PR to get a pre-merge test run

Squashed together @YashiKamboj's commits, added new commit for nits


fredemmott:
@YashiKamboj I'm removing this from the `all` list as it blames the `script` node - i.e. the error message does not point at the actual error. I'm happy to re-add that once it's addressed, but it will need changes to the `getNamespaces` thi9ng, so that will not be in a .z release

PR #329, opened 2020-11-03
Update schema for 4.82

fredemmott:
Should be backwards compatible


fredemmott:
Merging without macos test results, nothing MacOS-specific here

PR #331, opened 2020-11-20
Permit Missing `MarkupSuffix`

fredemmott:
This fixes current nightly builds; these always have a markup section
with an added hashbang token (even 0-width) and a missing `MarkupSuffix`

I manually edited the inferred_relationships file here to minimize the
codegen change; regenerating it properly removes the PU stuff, which
would turn this into a bigger BC breakage.


fredemmott:
heh, it doesn't like the manual code edit, which is fair enough it's hacky. I'll update it properly

jjergus (approved):
You could regenerate it with the HHVM version that generated the current one to avoid the BC breakage.

fredemmott:
Regenerating with the old version doesn't result in the necessary changes, as it doesn't parse anything in a way that results in a missing markupsuffix

fredemmott:
This should fail except on nightly, as new enough versions aren't released

fredemmott:
Going to come back to this soon.  I don't see why --rebuild-relationships isn't picking up that attribute type specifiers can have types other than ISimpleCreationSpecifier (things that can be implemented or extended); in this case, `<<__Soft>> dict<foo, bar>`. Once that's picked up and relationships are rebuilt, tests should pass.

PR #334, opened 2021-01-07
use shared github action

jjergus:
Output: https://github.com/hhvm/hhast/actions/runs/470034957

Everything's not ideal but this seems to be the best we can do with the current version of "composite action".

- We can't have multiple top-level sections from the same action (typecheck/lint/test) but at least I was able to add second-level grouping inside the action output.
- Not every part of the workflow configuration can be shared (e.g. the OS matrix) so there's still quite a bit of boilerplate left, but at least we can reuse the larger scripts (install HHVM/Composer).


fredemmott (approved):
> Not every part of the workflow configuration can be shared (e.g. the OS matrix)

I think this is actually desirable, and makes this easier to integrate with more complex build/test flows

PR #335, opened 2021-01-07
require HHVM 4.91 ?

jjergus:
4.89 and 4.90 was skipped, we probably want to require an existing HHVM version here?

On the other hand, there were several nightly releases tagged 4.89-dev which would also satisfy the original requirement -- maybe we want to preserve that? I'm not sure what's preferable.

Either way the release should probably be tagged as HHAST 4.91.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "hhvm/hacktest": "^2.2.0"
     },
     "require": {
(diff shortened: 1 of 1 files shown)


fredemmott (approved):
nice catch, and this matches the CI config

PR #339, opened 2021-01-20
Fix get_unresolved_referenced_names() for 4.93

fredemmott:


jjergus (approved):

codegen/syntax/TypeConstraint.hack
```diff
@@ -136,22 +136,22 @@ final class TypeConstraint extends Node {
...
   /**
-   * @return ClassnameTypeSpecifier | ClosureTypeSpecifier |
+   * @return ClassnameTypeSpecifier | ClosureTypeSpecifier | Contexts |
    * GenericTypeSpecifier | LikeTypeSpecifier | NullableTypeSpecifier |
    * ShapeTypeSpecifier | SimpleTypeSpecifier | TypeConstant |
    * VarrayTypeSpecifier | VectorTypeSpecifier
    */
-  public function getType(): ITypeSpecifier {
-    return TypeAssert\instance_of(ITypeSpecifier::class, $this->_type);
+  public function getType(): Node {
```
jjergus, line 144: Should we add `Contexts` to `ITypeSpecifier` too?

My plan was to create new interfaces for these (which I guess would mean also adding a new `wrap_IFoo` function for each) since neither `Context` nor `VariableToken` is really a type specifier, but now I'm not sure if that's worth the extra complexity.

codegen/syntax/ClassishBody.hack
```diff
@@ -158,25 +158,27 @@ final class ClassishBody extends Node {
 
   /**
    * @return NodeList<ConstDeclaration> | NodeList<IClassBodyDeclaration> |
+   * NodeList<ContextConstDeclaration> | NodeList<Node> |
```
jjergus, line 161: I think you can add `ContextConstDeclaration` to `IClassBodyDeclaration`. This one actually seems correct unlike the other 2.

jjergus (approved):

codegen/syntax/TypeConstraint.hack
fredemmott: `HackAstNodeTypeConstraint::getType()` returns an `IHackAstNodeSpecifierOrContexts` which is basically `ITypeSpecifier|Context` - I think we probably want to do something similar here :(

jjergus: Sounds good, that's pretty much exactly what I independently arrived at (except my interface name was `ITypeConstraintType`).

tests/TypeSpecifierOrContextsTest.hack
```diff
@@ -0,0 +1,51 @@
...
+      'unwrapped type specifier' => tuple(
+        'abstract const type TFoo as shape(...)',
+        ShapeTypeSpecifier::class,
+      ),
+      'wrapped type specifier' => tuple(
+        'abstract const type TFoo as MyClass',
+        SimpleTypeSpecifier::class,
+      ),
+      'contexts' =>
+        tuple('abstract const ctx CFoo super [defaults]', Node::class),
```
jjergus, line 29: should this be `Contexts::class`?

jjergus (approved):
lgtm

- don't forget to add `ContextConstDeclaration` to `IClassBodyDeclaration`
- do we want a similar interface (`ITypeSpecifierOrVariableToken`?) for the other one too?

fredemmott:
Current status:
- test pases for TypeConstraintsOrContexts. I expected it to fail as I haven't modified the deserialization code yet. Needs investigating
- I'm actually really tempted to throw that work away, and instead add a distinct `ContextsConstraint` HHAST node type that doesnt' exist in the FFP, and magically deserialize `type_specifier` into one or the other depending on context. These are really unrelated types.

tests/TypeSpecifierOrContextsTest.hack
fredemmott: whoops yep, but the SimpleTypeSpecifier one is the surprise

fredemmott: > don't forget to add ContextConstDeclaration to IClassBodyDeclaration

Done :)

> do we want a similar interface (ITypeSpecifierOrVariableToken?) for the other one too?


I don't think so: VariableTokens actually are *sometimes* specifying a type, similar to how a NameToken is *sometimes* specifying a type, depending on context. If it's not, you get a raw NameToken, if it is we generate a SimpleTypeSpecifier that doesn't exist in the CST

fredemmott:
> * I'm actually really tempted to throw that work away, and instead add a distinct `ContextsConstraint` HHAST node type that doesnt' exist in the FFP, and magically deserialize `type_specifier` into one or the other depending on context. These are really unrelated types.

I'm going to talk to Hunter about this; this feels like a better approach to me, and if it actually is we probably want to do this in HackAST too

fredemmott:
This is good to go: if we switch to the 'fake a ContextConstraint' class, for now that wouldn't be a breaking change as context constraints are currently a typechecker 1002 ('parse error', though the parser doesn't fail), so the affected syntax can't currently be used.

jjergus:
If everything passes then shipit, we can improve/refactor later -- but it would be good to have a working HHAST release tagged, so we can migrate everything else that depends on HHAST for linting.

fredemmott:
Doing a partial squash then merging, no actual changes between reviewed/test run and merge.

jjergus:
`ContextConstDeclaration` now has a `ContextConstraint` instead of `TypeConstraint` (https://github.com/facebook/hhvm/commit/bd819f75510f19b34e818356aafc2a176be3b9a2 -- apparently it was just a bug?)

That means we can probably delete `ITypeSpecifierOrContexts` now, or is there any reason to keep it?

PR #340, opened 2021-01-29
update schema (2021-01-22-0001)

jjergus:
The schema is backwards compatible, but the generated code isn't (there's a new argument in the middle for a bunch of constructors). Do we want to release that as 4.94.1 or 4.95.0?

### DIFF
--- a/codegen/inferred_relationships.hack
+++ b/codegen/inferred_relationships.hack
@@ -1,7 +1,7 @@
 /**
  * This file is generated. Do not modify it manually!
  *
- * @generated SignedSource<<4a1d222459067c2cf86706fd92475c44>>
+ * @generated SignedSource<<47eadaa50ec56f929e2a3f8f2f9f06ca>>
  */
 namespace Facebook\HHAST\__Private;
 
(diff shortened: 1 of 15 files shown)


fredemmott:
I think this needs to be 4.95.0; ParameterDeclaration is too likely to be used by some third-party linter outside of this repo

jjergus:
OK, I guess I'll wait until Monday then, so we don't accidentally release HHAST 4.95 that doesn't actually support HHVM 4.95...

fredemmott (approved):

fredemmott:
> so we don't accidentally release HHAST 4.95 that doesn't actually support HHVM 4.95...

If it required HHVM 4.95 in composer.json, I think this would actually be fine: if we immediately released a HHAST 4.95.1, we'd then be in a situation where no suported release of HHVM would ever have worked on 4.95.0.

With it being backwards-compatible, that's not the case: if we have to release a 4.95.1 and it's a BC break, we'll be in the uncomfortable situation where users of HHVM 4.94 will get HHAST 4.95.0, and we can't ship them a bugfix

One possible approach is to temporarily restrict this to HHVM 4.95 in composer.json, release HHAST 4.95.0, then once HHVM 4.95 is out, definitely release 4.95.1: if there's a syntax change, adapt to it, and if there isn't or if it isn't BC-breaking, widen the version constraint again

PR #342, opened 2021-02-06
add lambda support to Unused.*Linter

### LINKED ISSUE #341: UnusedParameter and UnusedVariable false positives in lambda default arguments
This sample code:

```Hack
function ex(
	int $i,
	(function(int): int) $fn = ($a) ==> {
		foo($a);
		return 5;
	},
): void {
	$fn($i);
}

function foo(int $_i): void {}
```

produces these errors:

```
Parameter is unused
  Linter: Facebook\HHAST\UnusedParameterLinter
  Location: /Users/greenberg/code/repro/unused_repro.hack:3:29
@@ -1,6 +1,6 @@
 function ex(
  int $i,
- (function(int): int) $fn = ($a) ==> {
+ (function(int): int) $fn = ($_a) ==> {
    foo($a);
    return 5;
  },

Would you like to apply this fix?
  [y]es/[n]o/yes to [a]ll/n[o] to all: Variable is unused
  Linter: Facebook\HHAST\UnusedVariableLinter
  Location: /Users/greenberg/code/repro/unused_repro.hack:4:6
  Code:
  >$a
```

Both are very confused about what to do with the lambda as a parameter default value. I think UnusedParameter needs to look for parameters that are children of the declaration rather than descendants.

jjergus:
Fixes #341

- improves some other cases (unused params/variables *inside* lambdas are caught)
- some false negatives still exist (using a shadowed param/variable inside a lambda also marks it as used in the enclosing function) -- probably fixable but I didn't want to do too much in one PR

### DIFF
--- a/src/Linters/UnusedParameterLinter.hack
+++ b/src/Linters/UnusedParameterLinter.hack
@@ -25,15 +25,26 @@ final class UnusedParameterLinter extends AutoFixingASTLinter {
       return null;
     }
 
-    $name = $node->getName();
-    if (!$name is VariableToken) {
-      return null;
-    }
-    if (Str\starts_with($name->getText(), '$_')) {
+    $name_node = $node->getName();
+    if (!$name_node is VariableToken) {
       return null;
     }
 
-    if ($functionish is FunctionDeclaration) {
+    $name = $name_node->getText();
+    if (Str\starts_with($name, '$_')) {
+      return null;
+    }
+
+    // If this is a parameter of a lambda function, we should be looking in the
+    // lambda's body, not the enclosing function/method's body.
+    $lambda =
+      $functionish->getClosestAncestorOfDescendantOfType<LambdaExpression>(
+        $node,
+      );
+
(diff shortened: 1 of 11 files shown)


jjergus:
@ryangreenberg I can't add you as a reviewer but you might also be a good perseon to review this.

src/Linters/UnusedParameterLinter.hack
```diff
@@ -25,15 +25,26 @@ final class UnusedParameterLinter extends AutoFixingASTLinter {
       return null;
     }
 
-    $name = $node->getName();
-    if (!$name is VariableToken) {
+    $name_node = $node->getName();
```
ryangreenberg, line 28: For a future change, I think this is roughly where we want to check that $node is a child of $functionish, not  any descendant to fix #341

ryangreenberg (approved):
Nice addition! I think your use of `traverse()` in a few places makes it easier to read and less dependent on knowing the exact hierarchy of the AST.

PR #344, opened 2021-02-24
No newline at start of control flow block

### LINKED ISSUE #111: Add linter banning newline as the first line of a block

lexidor:
Fixes #111

### DIFF
--- a/src/Linters/NoNewlineAtStartOfControlFlowBlockLinter.hack
+++ b/src/Linters/NoNewlineAtStartOfControlFlowBlockLinter.hack
@@ -9,7 +9,7 @@
 
 namespace Facebook\HHAST;
 
-use namespace HH\Lib\Str;
+use namespace HH\Lib\{C, Vec};
 
 class NoNewlineAtStartOfControlFlowBlockLinter extends AutoFixingASTLinter {
   const type TNode = IControlFlowStatement;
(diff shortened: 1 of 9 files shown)


src/Linters/NoNewlineAtStartOfControlFlowBlockLinter.hack
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    $left_brace = $body->getLeftBrace();
+    $next_token = $this->getAST()->getNextToken($left_brace);
+
+    if ($next_token is null) {
+      return null;
+    }
+
+    $whitespace = $next_token->getLeading()->getCode();
```
jjergus, line 35: This seems to rely on the assumption that the first EOL will be part of the left brace's trailing trivia and any other EOLs will be part of the next token's leading trivia.

I guess that's probably true right now but I'm not sure if we can rely on that never changing...it might be more robust to do something like this here:

```Hack
$whitespace =
  $left_brace->getTrailing()->getCode().
  $next_token->getLeading()->getCode();
```

(then check for double EOL at the beginning)

BTW do we have/should we have a linter for empty lines at the *end* of a control block?

src/Linters/NoNewlineAtStartOfControlFlowBlockLinter.hack
```diff
@@ -0,0 +1,78 @@
...
+      \get_class($node),
+    );
+  }
+
+  private static function fix(Token $token): Token {
+    return $token->withLeading(
+      $token->getLeading()->toVec()
+        |> new Vector($$)
+        |> $$->skipWhile($t ==> $t is EndOfLine)
+        |> new NodeList(vec($$)),
```
jjergus, line 75: nit: It might be better to avoid converting to a deprecated type (Vector) just so we can call one useful method -- maybe use a combination of HSL functions or just a loop?

jjergus (approved):
lgtm, just a few nitpicks

src/Linters/NoNewlineAtStartOfControlFlowBlockLinter.hack
lexidor: That would not work in case of
```HACK
if(true){//comment

  f();
}
```

If this hidden knowledge is subject to change, we can accept this false negative, right?

src/Linters/NoNewlineAtStartOfControlFlowBlockLinter.hack
lexidor: Note to self for later, find_key_with_value |> slice. I don't think the first one is standard hsl though. Raw loop if this is true.

src/Linters/NoNewlineAtStartOfControlFlowBlockLinter.hack
jjergus: I don't really have an opinion on whether that should count as a false positive/negative, but either way, it shouldn't be too hard to implement whichever exact check we decide on using either the concatenated getCode() strings, or the concatenated list of trailing+leading trivia.

Another edge case I can think of would be something like

```Hack
if (true) { /* multi-line
  comment */

  f();
}
```

and there might be others...

lexidor: I added these test cases. Current AST treats them correctly. If the AST changes and hands the eol to the next node, this continues to work. If all trivia is handed to the next node, only pure eol cases will continue to work. Those are the most common anyway. The new autofix doesn't force all code on the same line as the curly if the AST were to shift. I think this state is a good tradeoff between code complexity now and correct behavior later.

jjergus:
Seems reasonable, just one last thing: Can you extract the duplicate getBody function somewhere and reuse it?

jjergus:
Thanks!

Issue #269, opened 2020-02-19
Replace `classname<T>` usage with reified generics


fredemmott:
This should happen just as a modernization/tech debt issue, but no hurry - and it is a breaking change.

PR #345, opened 2021-02-25
Add new reified generic apis where classname<T> was previously used

lexidor:
Addresses #269 partially.
This doesn't replace the classname api with a reified one.
It supplements it with new one that are reified.
This way linters can remain hhast version agnostic.
The old api can be removed when the new ones have been around for a long time.
The `<<SoftDeprecated>>` was used instead of `<<__Deprecated>>`.
We can later grep for this and really `<<__Deprecated>>` them.

The first commit adds the new api and starts using it in the file itself.
The second commit converts as many uses as possible through the entire codebase.
The third commit is a realization that one of the deprecated methods was protected and in a final class.
It could be removed without breaking BC.

### DIFF
--- a/src/Linters/ASTLinter.hack
+++ b/src/Linters/ASTLinter.hack
@@ -9,14 +9,15 @@
 
 namespace Facebook\HHAST;
 
-use namespace Facebook\{HHAST, TypeAssert};
 use namespace Facebook\HHAST\SuppressASTLinter;
 use namespace HH\Lib\Vec;
 
 abstract class ASTLinter extends BaseLinter {
+  <<__Enforceable, __Reifiable>>
   abstract const type TContext as Node;
+  <<__Enforceable, __Reifiable>>
   abstract const type TNode as Node;
-  private ?HHAST\Script $ast;
+  private ?Script $ast;
 
   abstract protected function getLintErrorForNode(
     this::TContext $context,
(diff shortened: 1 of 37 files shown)


src/Linters/ASTLinter.hack
```diff
@@ -9,14 +9,15 @@
 
 namespace Facebook\HHAST;
 
-use namespace Facebook\{HHAST, TypeAssert};
 use namespace Facebook\HHAST\SuppressASTLinter;
 use namespace HH\Lib\Vec;
 
 abstract class ASTLinter extends BaseLinter {
+  <<__Enforceable, __Reifiable>>
```
lexidor, line 16: This is a breaking change.
You could previously target `ListItem<SomeNode>` without a typechecker error.
However, this was a bug imo, since the inner type was effectively ignored.
`ListItem<Node>` is a valid type which is no longer supported.

src/Linters/ASTLinter.hack
```diff
@@ -9,14 +9,15 @@
...
   abstract const type TContext as Node;
+  <<__Enforceable, __Reifiable>>
   abstract const type TNode as Node;
```
lexidor, line 19: This is a breaking change.
You could previously target `ListItem<SomeNode>` without a typechecker error.
However, this was a bug imo, since the inner type was effectively ignored.
`ListItem<Node>` is a valid type which is no longer supported.

PR #346, opened 2021-02-25
Delete linters that lint against invalid Hack code

lexidor:
MethodClassOnConstructor had been killed earlier.
It is already a Hack error:
Parsing[1002] Method call following immediate constructor call
requires parentheses around constructor call.
NoBasicAssignmentFunctionParameter checks for assignment as expression.
This is already a Hack error:
Parsing[1002] Assignments can no longer be used as expressions.
Pull the assignment out into a separate statement.


lexidor:
Is deleting a linter considered a BC break?
Configs mentioning the name are going to throw.
Users who disabled them in their configs using `"disabledLinters": [...]` for performance reasons are going to get hit.

jjergus:
> Is deleting a linter considered a BC break?

IMHO it's no worse than adding a linter, that's even more likely to require changes to any dependent projects (especially in this case where it only affects projects that explicitly mentioned these linters in their configs, which should be pretty rare).

PR #347, opened 2021-02-25
Make autofix suggestion of NoEmptyStatementsLinter safe (previously caused fatal error)

### LINKED ISSUE #146: NoEmptyStatementLinter suggests unsafe autofix for empty statements in braceless conditionals
The `NoEmptyStatementLinter` currently suggests removing empty statements. Here is an edge case where this is problematic.

```Hack
$x = 1;
$condition = false;

// explicitly do nothing in this case
if ($condition) ;

$x = 2;
echo $x;
```

In this case removing the semicolon causes a semantic change. Before `$x = 2` is executed unconditionally; after it is executed only when the condition is true. This is only a problem with braceless if statements.

A few options:

1. Never suggest autofixes for empty statements
2. Detect this case and do not suggest an autofix
3. Detect this case and suggest an autofix that adds empty braces
4. Detect this case and suggest removing the entire conditional statement

Having used this linter to automatically remove hundreds of stray semicolons, I don't like option 1. Option 2 is fine, and option 3 is only slightly more work. Option 4 seems pretty aggressive and more than is needed to fix this issue.

lexidor:
Inspired by https://github.com/hhvm/hhast/issues/146
Fixes: _Argument 5 passed to Facebook\HHAST\IfStatement::\_\_construct() must implement interface Facebook\HHAST\IStatement, Facebook\HHAST\NodeList given._
First commit demonstrates the failure.
This also fixes #146, since we replace the `;` with `{}` when dealing with control flow body_likes.

### DIFF
--- a/src/Linters/NoEmptyStatementsLinter.hack
+++ b/src/Linters/NoEmptyStatementsLinter.hack
@@ -9,8 +9,7 @@
 
 namespace Facebook\HHAST;
 
-use type Facebook\HHAST\{ExpressionStatement, Node, NodeList, Script, Token};
-
+use type Facebook\HHAST\{ExpressionStatement, Node, Script, Token};
 
 final class NoEmptyStatementsLinter extends AutoFixingASTLinter {
   const type TNode = ExpressionStatement;
@@ -43,7 +42,6 @@ final class NoEmptyStatementsLinter extends AutoFixingASTLinter {
         $this,
         'This statement includes an expression that has no effect',
         $stmt,
-        () ==> $this->getFixedNode($stmt),
       );
     }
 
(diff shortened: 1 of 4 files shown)


src/Linters/NoEmptyStatementsLinter.hack
```diff
@@ -43,21 +42,40 @@ final class NoEmptyStatementsLinter extends AutoFixingASTLinter {
         $this,
         'This statement includes an expression that has no effect',
         $stmt,
-        () ==> $this->getFixedNode($stmt),
```
lexidor, line 46: This removal doesn't change the behavior.
The autofix previously did an early return when the expression was null.
This just removes the autofix quick fix in the IDE, since it was not going to do anything anyway.

PR #349, opened 2021-02-26
Dont have two empty lines in a row linter

### LINKED ISSUE #109: Add linter banning consecutive blank lines

lexidor:
Fixes #109


src/Linters/suppress_ast_linter_error.hack
```diff
@@ -54,7 +54,11 @@ function is_linter_suppressed_in_sibling_node(
   string $fixme,
   string $ignore,
 ): bool {
-  $token = $root->getPreviousToken($node->getFirstTokenx());
```
lexidor, line 57: This threw an Exception, expected nonnull, got null.

src/Linters/DontHaveTwoEmptyLinesInARowLinter.hack
```diff
@@ -0,0 +1,79 @@
...
+          );
+        }
+      } else {
+        $eol_count = 0;
+      }
+    }
+
+    // We don't check the boundary with the next token.
+    // This will be checked by the loop above when linting
+    // that token instead.
```
jjergus, line 45: What if trailing ends with 2 or more empty lines? I don't know if this can happen.

src/Linters/DontHaveTwoEmptyLinesInARowLinter.hack
```diff
@@ -0,0 +1,79 @@
...
+    $eol_count = 0;
+    foreach ($token->getTrailing()->toVec() as $trivia) {
+      if ($trivia is EndOfLine) {
+        $eol_count++;
+        if ($eol_count === 3) {
+          return new ASTLintError(
+            $this,
+            "Don't have two empty lines in a row",
+            $token,
+            () ==> static::removeLeading($token, $trivia),
```
jjergus, line 55: Should this be `removeTrailing`? If possible, update the test to cover this branch (but I realize it might be impossible).

jjergus (changes requested):
Another linter against something I like doing :D But all the empty lines that were caught and removed here look like improvements, so it's probably good overall.

Can you update the test with:

1. add 3 or more subsequent empty lines somewhere (I think this case might not work correctly right now?)
2. add some comments, including a multi-line comment with 1,2,3+ empty lines (I'm not sure if we want to lint for empty lines in comments -- your choice/whatever is easier to implement -- but either way let's have the behavior tested)

Is performance a concern here? (there can be a lot of tokens in a file) Though I guess using EndOfLine as TNode limits that so hopefully should be OK (nice trick!)

lexidor:
> add some comments, including a multi-line comment with 1,2,3+ empty lines (I'm not sure if we want to lint for empty lines in comments -- your choice/whatever is easier to implement -- but either way let's have the behavior tested)

I test comments here:
https://github.com/hhvm/hhast/pull/349/files#diff-3871c84057c478d49774d8cc26040ef36c3177347a3b49d5df578b0ec4a27fa3R74

I don't touch comments, since their newlines are not EndOfLine nodes. A multiline comment is one big node instead.

lexidor:
> add 3 or more subsequent empty lines somewhere (I think this case might not work correctly right now?)

I can't emit multiple errors for one node sadly. So if you have five new lines, an error is emitted for the third. Once this one is deleted, another error is emitted on the next run. I will add a testcase for this, just in case.

jjergus:
Oh sorry I missed that, I was only looking at the .in file :)

lexidor:
Handling the case of multiple errors per node is not autofixable. I don't get handed the node to fix by hhast. So my autofix always operates on a stale node. If there are multiple errors on the same node, all autofixes except for the last one are effectively ignored. Thanks for pointing this out anyway. This has taught me valuable lessons about the HHAST internals.

jjergus:
> I can't emit multiple errors for one node sadly. So if you have five new lines, an error is emitted for the third. Once this one is deleted, another error is emitted on the next run. I will add a testcase for this, just in case.

If I'm reading it correctly, the node you're returning is a Token with its leading/trailing Trivia lists modified -- is it not possible to remove more than 1 item from the returned Token's Trivia list?

Alternatively, maybe this could be implemented at the TNode (EndOfLine) level, i.e. you'd return an autofix that replaces an EndOfLine node with `null`/removes it when it finds that it's preceded by more than one other EndOfLine node(s). This could be simpler and/or more performant, I'm not sure.

lexidor:
This fixing the parent node was done because I did not know a way to "delete" a node from the AST. Returning `null` is seen as a sign to not do anything. I can return an empty `new NodeList()`, which has the right effect on the textual representation of the autofixed code. However, other linters are not expecting to find a `NodeList` inside of a travia list. The _real_ way of fixing this is teaching the replacer a signal which means "I want to delete this node". I'll collect my thoughts and continue tomorrow.

jjergus:
Oh you're right, I didn't realize that.

IIRC we actually intentionally disallow removing nodes (there used to be methods like `$node->removeDescendant(...)` but no more) because it's impossible to make that type-safe. So you're right, you'll have to operate on the parent Token like you do now. Still, it should be possible to remove an arbitrary number of EndOfLines from the Token's Trivia lists instead of just one.

lexidor:
> Oh you're right, I didn't realize that.
> 
> IIRC we actually intentionally disallow removing nodes (there used to be methods like `$node->removeDescendant(...)` but no more) because it's impossible to make that type-safe. So you're right, you'll have to operate on the parent Token like you do now. Still, it should be possible to remove an arbitrary number of EndOfLines from the Token's Trivia lists instead of just one.

Sorry that this took me a couple tries to understand. I was staring right past the point you were trying to make all this time. I guess I have been staring at the AST for to long today 😆. Is this what you mean?

Instead of stopping at the first incorrect EOL, just fix all errors on the same node in one pass. Then remove all offending EOLs.

PR #350, opened 2021-02-26
Dont shadow local variables in foreach linter

### LINKED ISSUE #249: add linter against shadowing local variables in foreach
This should be a lint error:

```hack
$value = 42;
// ...
foreach ($arr as $value) {
  // ...
}
```

Same for function arguments:

```hack
function foo(int $value) {
  // ...
  foreach ($arr as $value) {
    // ...
  }
}
```

lexidor:
Fixes #249

### DIFF
--- a/src/Linters/DontShadowLocalVariablesInForeachLinter.hack
+++ b/src/Linters/DontShadowLocalVariablesInForeachLinter.hack
@@ -1,0 +1,164 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+use namespace HH\Lib\{C, Dict, Keyset, Str, Vec};
+
+final class DontShadowLocalVariablesInForeachLinter extends ASTLinter {
+  const type TContext = IFunctionishDeclaration;
+  const type TNode = ForeachStatement;
+
+  <<__Override>>
+  public function getLintErrorForNode(
+    this::TContext $scope,
+    this::TNode $foreach,
+  ): ?ASTLintError {
+    $pseudo_declarations = static::getPseudoDeclarations($foreach);
+    if (C\is_empty($pseudo_declarations)) {
+      return null;
+    }
+    $scope_decl = i_functionish_get_decl_header($scope);
+    $params = $scope_decl->getParameterList()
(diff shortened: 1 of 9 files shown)


src/Linters/DontShadowLocalVariablesInForeachLinter.hack
```diff
@@ -0,0 +1,164 @@
...
+    this::TContext $scope,
+    this::TNode $foreach,
+  ): ?ASTLintError {
+    $pseudo_declarations = static::getPseudoDeclarations($foreach);
+    if (C\is_empty($pseudo_declarations)) {
+      return null;
+    }
+    $scope_decl = i_functionish_get_decl_header($scope);
+    $params = $scope_decl->getParameterList()
+      ?->getDescendantsOfType(VariableToken::class)
```
jjergus, line 29: This can be incorrect if any of the parameters has a lambda function as its default value (I recently had to fix a similar bug in https://github.com/hhvm/hhast/pull/342, you can look at the test cases there -- maybe copy some of them here too).

It should probably only look at direct children, not all descendants.

You might also be able to extract and reuse some code from UnusedVariableLinter, e.g. the parts where it computes the list of params from a lambda/non-lambda function (getLambdaParts, getFunctionishParts).

src/Linters/DontShadowLocalVariablesInForeachLinter.hack
```diff
@@ -0,0 +1,164 @@
...
+    return Vec\map(
+      $lambdas,
+      $lambda ==>
+        $lambda->getSignature()->getDescendantsOfType(VariableToken::class),
+    )
+      |> Vec\flatten($$)
+      |> Keyset\map($$, $p ==> $p->getText());
+  }
+
+  private static function getAssignedLocalsBeforeForeach(
```
jjergus, line 104: This looks potentially similar to UnusedVariableLinter::classifyVariables, can we extract/reuse it?

jjergus (changes requested):
Looks good but can you try reusing as much as possible from `UnusedParameterLinter` and `UnusedVariableLinter`? (those two are likely already duplicating stuff -- feel free to fix anything you run into if you have time)

Reusing the code might fix some bugs (see inline), but also maybe you'll find new bugs affecting `UnusedParameterLinter` or `UnusedVariableLinter` that your code doesn't have -- hopefully all 3 linters will end up more robust as a result if all their core parts are shared.

lexidor:
Definitely worth drying this code up. A lot of concepts are indeed neigh identical.

lexidor:
This is already included in HHClientLinter.

PR #351, opened 2021-02-26
Support multiple forms of inline suppression

### LINKED ISSUE #308: Support multiple forms of inline suppression
HHAST_FIXME and HHAST_IGNORE_ERROR were originally intended to mirror HH_FIXME vs HH_IGNORE_ERROR

- HH_IGNORE_ERROR has been removed
- the 'error' part makes some people feel more uncomfortable suppressing linters than they should; linters should generally be taken as 'this is fishy', not a hard error. There are valid reasons to e.g. await in a loop, and suppressing the linter is better/clearer than refactoring to avoid triggering it

Linters can provide their own override, but that replaces the default set.

I suggest:
1. changing this to a `keyset<string>` so that specific linters can *add* a supported suppression
2. adding a few more defaults:
  - `HHAST_IGNORE[Classname]`: still consistent, but briefer, and doesn't have the negative connotations of FIXME or ERROR
  - `@lint-ignore ClassName`: consistent with some other frameworks, and still a low chance of collision - and it doesn't matter if there is a collision
3. (optional) allow projects to configure their own additional suppression comments for each linter
 - for example, could support `@lint-ignore FacebookInternalLinterName` in the HSL

lexidor:
Fixes #308
This PR adds the following default suppressors.
@lint-ignore x, @lint-fixme x, @lint-ignore-all x, HHAST_IGNORE[x].
You may now also specify suppression aliases in hhast-lint.json.

### DIFF
--- a/hhast-lint.json
+++ b/hhast-lint.json
@@ -22,5 +22,8 @@
         "Facebook\\HHAST\\DataProviderTypesLinter"
       ]
     }
-  ]
+  ],
+  "suppressionAliases": {
+    "Facebook\\HHAST\\DontAwaitInALoopLinter": [ "This is a polling loop" ]
+  }
 }
--- a/src/Linters/BaseLinter.hack
+++ b/src/Linters/BaseLinter.hack
@@ -11,7 +11,7 @@ namespace Facebook\HHAST;
 
 use namespace Facebook\TypeAssert;
 use type Facebook\HHAST\File;
-use namespace HH\Lib\{C, Str};
+use namespace HH\Lib\{C, Keyset, Str};
 
 <<__ConsistentConstruct>>
 abstract class BaseLinter {
@@ -27,6 +27,7 @@ abstract class BaseLinter {
   public function __construct(
     private File $file,
     private ?this::TConfig $config,
+    private keyset<string> $extraSuppressors,
   ) {
   }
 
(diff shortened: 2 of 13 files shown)


jjergus:
I have mixed feelings about this...what's the value of having so many different keywords?

The value of having a single keyword is that you can do `grep HHAST_IGNORE_ERROR *.hack` to find all ignored lints, there's less stuff for people to remember when writing/reading code, etc.

lexidor:
I think adding `HHAST_IGNORE` is a good change. `HHAST_IGNORE_ERROR` suggests the linter away behavior is erroneous. The `@lint-*` are meant for interoperability with HackAST. This allows Facebook to lint against the hsl directory. I guessed HackAST's syntaxes for global and fixme.

I don't know for sure if the alias feature I implemented matches what Fred had in mind. Should aliases only be able to alias the linter name, whilst still being required to have a `HHAST_*` or `@lint-*` prefix? I am starting to second guess supporting `// This is a polling loop`. It is unclear that it is suppressing a linter and looks like a useless inline comment. Wide use of this might discourage removal of useless inline comments, because they might be suppressing a linter.

fredemmott:
- HHAST_IGNORE seems like a good change, and has minor change to grepability
- for the others, perhaps it would be best to make it configurable via hhast-lint.json

lexidor:
Major merge conflict with refactored code base.

PR #352, opened 2021-02-27
Make var_dump on a node more readable

lexidor:
Working with `\var_dump()` in hhast doesn't work most of the time.
Nodes often refer to themselves via another object.
This causes `var_dump()` to print the recursion until the recursion limit is reached.

Also add `->toDebugTree()`, which has the same contents are `__debugInfo()`,
but you can use IDE features to expand sections and walk the tree (see example).

### DIFF
--- a/src/nodes/Node.hack
+++ b/src/nodes/Node.hack
@@ -361,4 +361,31 @@ abstract class Node implements IMemoizeParam {
     }
     return null;
   }
+
+  public function __debugInfo(): dict<arraykey, mixed> {
+    return Dict\merge(
+      dict[
+        '__ID__' => $this->getUniqueID(),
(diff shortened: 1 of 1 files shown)


lexidor:
```HACK
namespace Facebook\HHAST;
<<__EntryPoint>>
async function inspect_myself_async(): Awaitable<void> {
  require_once __DIR__.'/vendor/autoload.hack';
  \Facebook\AutoloadMap\initialize();

  $self = await from_file_async(File::fromPath(__FILE__));
  echo \json_encode(
    $self->getDescendantsByType<FunctionCallExpression>()[3]->toDebugTree(),
    \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES,
  );
}
```

image

PR #355, opened 2021-03-13
Allow MemberSelectionExpressions as defaults in function headers

### LINKED ISSUE #354: UnusedVariable false positive in method signature
This sample code:

```hack
<?hh

final class C {
	private int $x = 1;
	public function say(int $y = $this->x): void {
		echo $y;
	}
}
```

Produces this error:

```
Variable is unused
  Linter: Facebook\HHAST\UnusedVariableLinter
  Location: tests/test.php:5:30
  Code:
  >$this
```

It seems like the linter should not view `$this` as a variable, since we're just accessing the `$x` property.

I'm running HHVM 4.99 and using the latest version of HHAST as well.

ianhoffman:
Fixes #354.

The fix I came up with is just to bail out immediately if the variable expression is part of a `MemberSelectionExpression`, as these can never result in unused variables. For example:

* `$this->x = 1`. All variables are used.
* `$x = $this->y` .`$this` is used. `$x` will be linted separately.

I don't love introducing a special case for `MemberSelectionExpression`, however, and I'm wondering how else I might approach this. This is my first time hacking on HHAST so suggestions are very welcome.

Thanks!

### DIFF
--- a/src/Linters/UnusedVariableLinter.hack
+++ b/src/Linters/UnusedVariableLinter.hack
@@ -21,6 +21,11 @@ final class UnusedVariableLinter extends AutoFixingASTLinter {
     VariableExpression $node,
   ): ?ASTLintError {
 
+    $parent = $functionish->getParentOfDescendant($node);
+    if ($parent is MemberSelectionExpression) {
+      return null;
+    }
+
     $var = $node->getExpression();
     if (!$var is VariableToken) {
       return null;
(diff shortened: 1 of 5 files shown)


src/Linters/UnusedVariableLinter.hack
```diff
@@ -21,6 +21,11 @@ final class UnusedVariableLinter extends AutoFixingASTLinter {
     VariableExpression $node,
   ): ?ASTLintError {
 
+    $parent = $functionish->getParentOfDescendant($node);
+    if ($parent is MemberSelectionExpression) {
```
fredemmott, line 25: This is probably the best way for now. At some point it would probably be worth adding a quick "don't recurse into this subtree" interface to `ASTLinter` just for performance, but this is fine.

Could you copy the explanation from your PR into inline code comments here please?

fredemmott (approved):

fredemmott:
Actually, I'll merge it then copy your description :)

ianhoffman:
> Actually, I'll merge it then copy your description :)

Thanks! 🙇‍♂️

PR #356, opened 2021-03-17
Centralize parser cache files in `$REPO_ROOT/.var`

### LINKED ISSUE #254: Unwanted cache files created in VS Code
I just launched VS Code in a folder with my hack files, and I got a dialog about using hhast. I confirmed it and in a sec it created bunch of files, which are now "messing" up my repo:
Screenshot 2020-02-05 18 19 46
Is this expected? Can we store these somewhere else?

fredemmott:
fixes #254

Having the cache files right next to their source has a few problems:
- it's slightly annoying to have to tell the IDE to ignore these if IDE
doesn't respect .gitignore
- it's problematic for other standard tools like `find -type f src/`

This uses `REPO_ROOT/.var/cache/hhvm/hhast/PATH_SHA` instead:
- this fixes the above problems
- the idea is `REPO_ROOT/.var` is used similarly to `/var` on Linux
  systems:
  - other tools (e.g. autoloader) could store things here without
    needing another .gitignore entry
  - if we end up wanting to make hhast have long-running-but-attachable
    processes, `REPO_ROOT/.var/run/hhvm/hhast/pid` would be the obvious
    choice
  - it might make sense to modify the typechecker to use this subdir
    instead of `/tmp/hh_server/`, which has several problems on
    multi-user systems

Also, I'm using relative paths for PATH_SHA; this allows priming the
cache from some other script (e.g. a checkout script or dev environment
setup).


src/__Private/ParserCache.hack
```diff
@@ -71,11 +72,47 @@ final class ParserCache {
...
+
+  private function getCacheFileName(File $file): ?string {
     if ($file->isDirty()) {
       return null;
     }
+
     $source = $file->getPath();
+    $root = $this->getRepoRoot($source);
+    if ($root !== null) {
+      $dir = $root.'/.var/cache/hhvm/hhast';
```
jjergus, line 109: Maybe just `$root.'/.var/hhast-cache'`? Or do you expect each of the directories to contain other stuff in the future so the multi-level hierarchy is warranted?

jjergus (approved):
I don't have a strong opinion but why not use the path of each file instead of the hash? (e.g. `src/foo/Bar.hack` --> `.var/hhast-cache/src/foo/Bar.hack.parser-cache`)

It could make debugging etc. easier.

src/__Private/ParserCache.hack
fredemmott: I was intending to use github orgs as a vendor ID to avoid conflicts

PR #357, opened 2021-03-20
Fix MustUseBracesForControlFlowLinter autofix

lexidor:
for and do were checked, but could not be fixed.
This resulted in an invariant exception.

### DIFF
--- a/src/Linters/MustUseBracesForControlFlowLinter.hack
+++ b/src/Linters/MustUseBracesForControlFlowLinter.hack
@@ -42,24 +42,17 @@ class MustUseBracesForControlFlowLinter extends AutoFixingASTLinter {
   }
 
   private function getLastHeadToken(Node $node): Token {
-    if ($node is IfStatement) {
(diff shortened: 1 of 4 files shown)


src/Linters/MustUseBracesForControlFlowLinter.hack
```diff
@@ -42,24 +42,17 @@ class MustUseBracesForControlFlowLinter extends AutoFixingASTLinter {
   }
 
   private function getLastHeadToken(Node $node): Token {
-    if ($node is IfStatement) {
-      $paren = $node->getRightParen();
-      if ($paren !== null) {
-        return $paren->getLastTokenx();
-      }
-      return $node->getCondition()->getLastTokenx();
```
jjergus, line 50: Do we need to keep this special case for IfStatement?

jjergus: I guess not, I see now that `IfStatement::getRightParen()` is not nullable.

PR #359, opened 2021-04-20
New attribute syntax parser support

lexidor:
The codegen doesn't know about the new attribute syntax just yet.
`must be of type ?Facebook\HHAST\OldAttributeSpecification, Facebook\HHAST\AttributeSpecification given`
We must give examples of the new attribute syntax to hhast and rebuild the codegen.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@ tests/lsp/*.out
 *.swo
 .DS_Store
 .var/
+*.hhast.parser-cache
 composer.lock
--- a/.hhconfig
+++ b/.hhconfig
@@ -3,3 +3,5 @@ error_php_lambdas = true
 disable_xhp_children_declarations = false
 allowed_decl_fixme_codes=2053,4045,4047
 allowed_fixme_codes_strict=2011,2049,2050,2053,2083,3084,4027,4045,4047,4104,4106,4107,4108,4110,4128,4135,4188,4223,4240,4323,4390,4401
+disallow_silence = true
+allow_new_attribute_syntax = true
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,5 +3,6 @@
   "hack.useHhast": true,
   "files.associations": {
     "*.php": "hack"
-  }
+  },
+  "editor.tabSize": 2
 }
--- a/src/__Private/ParserCache.hack
+++ b/src/__Private/ParserCache.hack
@@ -22,6 +22,7 @@ final class ParserCache {
   const type TCached = shape(
     'hash' => string,
     'ast' => dict<string, mixed>,
+    ?'new_attribute_syntax_allowed' => bool,
   );
 
   public function fetch(File $file): ?dict<string, mixed> {
@@ -35,7 +36,7 @@ final class ParserCache {
 
     try {
       $cache = \unserialize(\gzinflate(\file_get_contents($path)) as string)
-        |> TypeAssert\matches_type_structure(self::getCacheTS(), $$);
+        |> TypeAssert\matches<self::TCached>($$);
     } catch (\Exception $_) {
       \unlink($path);
       return null;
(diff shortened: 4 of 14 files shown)


lexidor:
I just codegenned the new attributes codegen.
It is not possible to support `$resource = @fopen(__FILE__);` and `@__Memoize() function life(): int { return 42; }` at the same time.
Types which mentioned `OldAttributeSpecification` before have become `Node`.
```DIFF
-  public function getAttribute(): ?OldAttributeSpecification {
+  public function getAttribute(): ?Node {
--
-  public function getAttributeSpec(): ?OldAttributeSpecification {
+  public function getAttributeSpec(): ?Node {
-  public function getAttributeSpecx(): OldAttributeSpecification {
+  public function getAttributeSpecx(): Node {
--
-  public function getAttributex(): OldAttributeSpecification {
+  public function getAttributex(): Node {
```

I'd love to know:
 - how to make both `AttributeSpecification` and `OldAttributeSpecification` implement a common interface.
 - how to decay the types to that interface instead of `Node`.
 - if error suppression support (`@fopen(__FILE__)`) can be dropped safely.
 - how to support linters that hardcode against `OldAttributeSpecification`.
 
In the meanwhile, this PR is safe to merge.
It will not fail to parse a file that previously parsed.
If will not parse a file with new attribute syntax either (since the types refer to `OldAttributeSpecification` still).

jjergus:
> how to make both AttributeSpecification and OldAttributeSpecification implement a common interface.

Manually create a new interface (either empty or with whatever methods we know to exist on all subclasses), then add it to `CodegenBase::getMarkerInterfacesByInterface()`.

> how to decay the types to that interface instead of Node.

should happen automatically after the previous step

> if error suppression support (@fopen(__FILE__)) can be dropped safely.
> how to support linters that hardcode against OldAttributeSpecification.

Is it possible to support both? Tests could be conditionally skipped and linters could conditionally do nothing based on ini_get (like you did in NewAttributeParsingTest.hack)

lexidor:
> Is it possible to support both? Tests could be conditionally skipped and linters could conditionally do nothing based on ini_get (like you did in NewAttributeParsingTest.hack)

The hard part is the codegen. The syntaxes are mutually exclusive. When using hh_parse without new attribute syntax support, the following code is parsed as:
```HACK
// error suppression operator + function name fopen +
// open paren + constant + close paren + semicolon
@fopen(__FILE__);
// error suppression operator + function name attr +
// open paren + close paren +
// invoke the return value function call +
// KABOOM (unexpected lambda arrow)
@attr() () ==> 42;
```

However, the new syntax explodes on the other one:
```HACK
// attribute start, attribute name fopen, open paren +
// constant + close paren +
// KABOOM (expected an lambda to annotate, got semicolon)
@fopen(__FILE__);
// attribute start + attribute name attr + open paren +
// close paren + empty lambda argument list +
// lambda arrow + lambda expression body + semicolon
@attr() () ==> 42;
```

@hgoldstein did I explain this correctly? You chimed in on facebook/hhvm#8829 and this is very much the same conundrum. This is why error suppression and new attributes are mutually exclusive, right?

This means that any file which uses the error suppression operator can not be parsed. The codegen will never see a syntax example with error suppression. It will "forget" to codegen the class for the error suppression operator. So we would have to complicate the codegen (2 passes one with new syntax and one without). We would have to merge the results. The thought of this scares me.

hgoldstein:
@lexidor: yes, having error suppression is mutually exclusive with the new attribute syntax as you can't disambiguate the two at parse time.

jjergus:
I agree we want this change eventually, but I'm not sure what the right time is. I think it depends on

1. How much do you need it? (is it blocking you from using HHAST in your codebase, etc.)
2. Would it cause problems to other HHAST users who may still be using @ for error supression?

lexidor:
I use hhast on every project I have some say in. I would not drop hhast to gain the new attribute syntax. There is admittedly not much to gain from the new attribute syntax. I would just like to adopt the new syntax at some point, since hackfmt does a consistent job at formatting them pleasantly. Instead of keeping attributes on one line and wrapping when it exceeds the line length, it always stacks them up. Hhast is the only thing holding me back from adopting it.

```HACK
<<__Memoize, __Override>>
public function...

<<
  LongNameToForceALineWrap,
  AndSomethingElseLong,
  ExampleNameWithManyCharacters,
>>
public function...

// Pleasantly self consistent, no matter the length
@__Memoize
@__Override
public function...
```

I'd love to know if the old hsl approach `PHPErrorSuppressor` would work for users who currently use the at sign syntax for suppression of the error handler.

lexidor:
Let's not fragment the community and wait for the moment where hhvm switches this on by default.

PR #361, opened 2021-04-20
Use type assertions instead of hh fixmes

lexidor:
Let's take a swing at these suppression comments.
There are tons and tons of them and most of them could be enforced.
AIUI, this doesn't degrade runtime performance.
Instead of doing the check at the runtime call `\TypeError` or `fatal`, it is just an `as` assertion.
If this is slower, I don't understand hhvm as well as thought I did.

Please don't merge my codegen though.
My codegen is always slightly off what everyone else generates.
Maybe ini settings /shrug/?
The codegen is included in the PR, so you can see what the changes I made are affecting.
Before merging, revert the codegen.


jjergus:
> AIUI, this doesn't degrade runtime performance.
Instead of doing the check at the runtime call \TypeError or fatal, it is just an as assertion.
If this is slower, I don't understand hhvm as well as thought I did.

I would expect this to degrade performance slightly (the runtime now has to do 2 checks, first when evaluating the newly added `as` expressions, then the normal function call param validation stuff) but the cost is likely very small, and there's a lot of value in removing FIXMEs, so this seems good to me.

lexidor:
> the runtime now has to do 2 checks, first when evaluating the newly added as expressions, then the normal function call param validation stuff

I had always assumed that either the jit or repoauth would realize that the type hint could never fail and kicked out the bytecode(s) that do this check.

jjergus:
I'm sure it gets optimized in some circumstances but I wouldn't 100% rely on it (does HHAST even run in repomode normally?). But again, I don't think it matters -- replacing FIXMEs with cheap checks is still good even if it's not completely free.

lexidor:
My codegen is always subtly different, so I'll redo the codegen in a docker container. Is it preferable to use the exact same hhvm version that the current codegen was run on? Or should I use the latest nightly? If we want the exact same version, I'll need to know which nightly was used here: 53600a0

jjergus:
If the schema changes are just comments (like in https://github.com/hhvm/hhast/pull/361/commits/fc12e0d3222346edc293dae6df0fb1a687101dad) I wouldn't worry about it.

lexidor:
> If the schema changes are just comments (like in [fc12e0d](https://github.com/hhvm/hhast/commit/fc12e0d3222346edc293dae6df0fb1a687101dad)) I wouldn't worry about it.

Well that, and the hhvm version id constant does not have a dev suffix. I guess it won'tnneed iit either, since 4.106 is already out.

PR #364, opened 2021-09-13
Support current nightlies

fredemmott:
- removes partially abstract type constant for linter config
- sets TConfig to `shape()` for all linters without a configuration
- removes test asserting that it's fine to provide a configuration to a
  linter that doesn't want one.

RE that last point: if a configuration changes the behavior of a linter,
and the user needs that configuration, it's probably not safe to run with
multiple versions of HHAST anyway.

### DIFF
--- a/src/__Private/LintRunConfig.hack
+++ b/src/__Private/LintRunConfig.hack
@@ -53,7 +53,7 @@ final class LintRunConfig {
     >,
     // Each linter may specify a type for itself.
     // The type of this key is effectively `dict<classname<T1 ... Tn>, Tx>`
-    ?'linterConfigs' => dict<string, BaseLinter::TConfig>,
+    ?'linterConfigs' => dict<string, dynamic>,
   );
 
   const type TFileConfig = shape(
--- a/tests/LinterConfigTest.hack
+++ b/tests/LinterConfigTest.hack
@@ -13,7 +13,6 @@ use type Facebook\HackTest\HackTest;
 use function Facebook\FBExpect\expect;
 use type Facebook\HHAST\{
   BaseLinter,
-  FinalOrAbstractClassLinter,
   LintError,
   PreferRequireOnceLinter,
 };
--- a/tests/LinterConfigTest.hack
+++ b/tests/LinterConfigTest.hack
@@ -108,18 +108,6 @@ final class LinterConfigTest extends HackTest {
     expect($config)->toBeNull('No config supplied');
   }
 
-  public function testConfigSuppliedAndLinterDoesNotSpecify(): void {
-    $lrc = static::getLintRunConfig();
-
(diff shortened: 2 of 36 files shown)


tests/LinterConfigTest.hack
```diff
@@ -108,18 +108,6 @@ final class LinterConfigTest extends HackTest {
...
-
-    $config = $lrc->getLinterConfigForLinter<FinalOrAbstractClassLinter, _>(
-      FinalOrAbstractClassLinter::class,
-    );
-
-    expect($config)->toEqual(
-      shape('I can supply a config' => "even if the linter doesn't need one"),
-    );
-  }
-
```
fredemmott, line 122: @lexidor  any thoughts on this change? The main alternative is to specify all linters as taking `shape(...)` but that removes a lot of the safety.

src/__Private/LintRunConfig.hack
```diff
@@ -53,7 +53,7 @@ final class LintRunConfig {
     >,
     // Each linter may specify a type for itself.
     // The type of this key is effectively `dict<classname<T1 ... Tn>, Tx>`
-    ?'linterConfigs' => dict<string, BaseLinter::TConfig>,
+    ?'linterConfigs' => dict<string, dynamic>,
```
lexidor, line 56: Why do we need dynamic here? Would `mixed` or `shape(...)` cause type errors?

tests/LinterConfigTest.hack
lexidor: When I originally wrote this, I imagined that many linters that were not configurable before might want to become configurable. I wanted adding a configuration to be a forward compatible change. This didn't really happen, so it would be safe to remove this.

src/__Private/LintRunConfig.hack
fredemmott: - `shape(...)` states that any shape is acceptable
- `mixed` indicates that any value is acceptable
- `dynamic` means that the acceptable types vary - and given we refine it at runtime, seems the most correct

lexidor: I have very little experience with the local permissive operations of dynamic. If this type was chosen for documentation ppurposes, would it be an idea to declare `type TConfigurationDiffersPerLinter = mixed;` and stick that in there. Then you can'tgget caught of guard by the typechecker not warning you about incorrect local operations, but the documentation purpose is very clear.

lexidor (approved):
✔️ Don't forget that this is a BC break for people who've written their own linters.

lexidor (approved):
✔️ Don't forget that this is a small BC break for people who've written their own linters.

src/__Private/LintRunConfig.hack
fredemmott: The local behavior shouldn't be an issue here, as it gets handled within the internals of hhast, not the linters

PR #367, opened 2021-10-25
Add simplified hhast-dump utility

fredemmott:
This is primarily for debugging HHAST itself, for cross-checking with
`hh_parse --full-fidelity-json` - for example, to distinguish between missing and 0-width nodes.

hh_parse remains the recommended way to canonically verify the parser's
behavior.

hhast-inspect (https://github.com/hhvm/hhast-inspect) remains the
recommended way to explore the AST of large files.

This distinction is useful when inspecting

### DIFF
--- a/bin/hhast-dump.hack
+++ b/bin/hhast-dump.hack
@@ -1,0 +1,40 @@
+#!/usr/bin/env hhvm
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST\__Private;
+
+<<__EntryPoint>>
+async function hhast_dump_main_async(): Awaitable<noreturn> {
+  $root = \realpath(__DIR__.'/..');
+  $found_autoloader = false;
+  while (true) {
+    $autoloader = $root.'/vendor/autoload.hack';
+    if (\file_exists($autoloader)) {
+      $found_autoloader = true;
+      require_once($autoloader);
+      \Facebook\AutoloadMap\initialize();
+      break;
+    }
+    if ($root === '') {
+      break;
+    }
+    $parts = \explode('/', $root);
+    \array_pop(inout $parts);
+    $root = \implode('/', $parts);
+  }
+
+  if (!$found_autoloader) {
+    \fprintf(\STDERR, "Failed to find autoloader.\n");
+    exit(1);
+  }
+
+  $result = await DumpCLI::runAsync();
+  exit($result);
+}
--- a/src/__Private/DumpCLI.hack
+++ b/src/__Private/DumpCLI.hack
@@ -1,0 +1,80 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST\__Private;
+use namespace Facebook\HHAST;
+
+use namespace HH\Lib\{C, Dict, Str};
+use type Facebook\CLILib\CLIWithRequiredArguments;
+use namespace Facebook\CLILib\CLIOptions;
+
+final class DumpCLI extends CLIWithRequiredArguments {
+  <<__Override>>
+  public static function getHelpTextForRequiredArguments(): vec<string> {
+    return vec['PATH'];
+  }
+
+  <<__Override>>
+  protected function getSupportedOptions(): vec<CLIOptions\CLIOption> {
+    return vec[];
+  }
+
+  <<__Override>>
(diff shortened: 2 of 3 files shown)


bin/hhast-dump.hack
```diff
@@ -0,0 +1,40 @@
...
+
+<<__EntryPoint>>
+async function hhast_dump_main_async(): Awaitable<noreturn> {
+  $root = \realpath(__DIR__.'/..');
+  $found_autoloader = false;
+  while (true) {
+    $autoloader = $root.'/vendor/autoload.hack';
+    if (\file_exists($autoloader)) {
+      $found_autoloader = true;
+      require_once($autoloader);
```
Atry, line 21: What is autoloader?

Atry (changes requested):
Could you rebase it onto the devcontainer, so I could test it in a Codespace?

bin/hhast-dump.hack
fredemmott: "You want class Foo? It's in src/foo.hack. Function bar? src/bar.hack"

An autoloader is required for all hack programs that are not single-file.

src/__Private/DumpCLI.hack
```diff
@@ -0,0 +1,80 @@
...
+    return $this->inlineOrIndent($this->textForNode($node));
+  }
+
+  private function textForNode(HHAST\Node $node): string {
+    $class = \get_class($node) |> Str\split($$, '\\') |> C\lastx($$);
+
+    if ($node is HHAST\Token) {
+      return vec[
+        $class.':',
+        '  leading:'.$this->inlineOrIndentNode($node->getLeading()),
```
Atry, line 63: Should we also indent `leading`/`text`/`trailing` sections?

src/__Private/DumpCLI.hack
```diff
@@ -0,0 +1,80 @@
...
+    await $this->getStdout()->writeAllAsync($this->textForNode($ast)."\n");
+    return 0;
+  }
+
+  private function inlineOrIndent(string $text): string {
+    $indent = '  ';
+    if (Str\length($text) < 60 && !Str\contains($text, "\n") && !Str\starts_with($text, '  ')) {
+      return ' '.$text;
+    }
+    return "\n".$indent.Str\replace($text, "\n", "\n".$indent);
```
Atry, line 50: The approach to indent is smart but less efficient, because the string replacing is performed repeatedly.

Atry (changes requested):
1. Need tests
2. Instead of formatting the output manually, shall we simply convert the AST to a tree of `shape`s, and print it as JSON, so that we don't have to maintain the code for pretty-printing.

src/__Private/DumpCLI.hack
fredemmott: debug tool, performance isn't a significant concern.

fredemmott:
One disadvantage of JSON is that there will be *no* indentation for multi-line continuations, but they're rare.

Atry:
> One disadvantage of JSON is that there will be _no_ indentation for multi-line continuations, but they're rare.

There is a [JSON_PRETTY_PRINT](https://github.com/facebook/hhvm/blob/3eed677a4c204220dd24b2c2c2c26fdf4e7c17a9/hphp/hack/hhi/stdlib/builtins_json.hhi#L39) flag

fredemmott:
yep, but that just gives structual indentation, not content-aware.

src/__Private/DumpCLI.hack
fredemmott: yes, and they are, they need to show up as sub-items.

fredemmott:
Re-requesting review:

- json is significantly less readable - e.g. https://gist.github.com/58ee2e0f44c807035b5f27943877c690
- tests will either be basically copy-pasting the implementation, or extremley brittle

Atry:
> I'll give JSON a try; if that is readable output, I'll add tests, otherwise I won't: it's a debug tool, and priority is to be human-readable, nto machine-readable, so any meaningful tests would be brittle and extremely coupled to the parser.

I understand it would be brittle when the AST changes, but the burden to maintain some tests are not super high as long as the the expected output can be generated and updated. It is actually an opportunity to maintain a set of snapshot test cases, making any changes in AST be visible to the PR reviewer.

fredemmott:
> It is actually an opportunity to maintain a set of snapshot test cases, making any changes in AST be visible to the PR reviewer.

This is covered in Hack's own tests, and by the changes to codegen/

fredemmott:
> making any changes in AST be visible to the PR reviewer.

We should also intentionally hide anythign tightly coupled to the AST from reviewers with @generated :p

Atry:
> Re-requesting review:
> 
> * json is significantly less readable - e.g. https://gist.github.com/58ee2e0f44c807035b5f27943877c690
> * tests will either be basically copy-pasting the implementation, or extremley brittle

Makes sense. Accepting now.

Atry (approved):

fredemmott:
For reference, that was generated with https://gist.github.com/fredemmott/215f623abbb373289dc9d37f2119fb0d

PR #368, opened 2021-11-04
Add HHClientLinter

Atry:
This is the initial implementation of `HHClientLinter`, which is a proxy calling `hh_client --lint`.

### DIFF
--- a/src/Linters/HHClientLintError.hack
+++ b/src/Linters/HHClientLintError.hack
@@ -1,0 +1,66 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+final class HHClientLintError implements LintError {
+
+  const type TJSONError = shape(
+    'descr' => string,
+    'severity' => HHClientLintSeverity,
+    'path' => string,
+    'line' => int,
+    'start' => int,
+    'end' => int,
+    'code' => int,
+    'bypass_changed_lines' => bool,
+    'original' => string,
+    'replacement' => string,
+  );
+
+  public function __construct(
+    private File $file,
+    private this::TJSONError $error,
+    private ?string $blame_code = null
+  ) {
+  }
+
+  public function getFile(): File {
+    return $this->file;
+  }
+
+  public function getDescription(): string {
+    return $this->error['descr'];
+  }
+
+  public function getPosition(): (int, int) {
+    return tuple($this->error['line'], $this->error['end']);
+  }
+
+  public function getRange(): ((int, int), (int, int)) {
+    return tuple(
+      tuple($this->error['line'], $this->error['start']),
+      tuple($this->error['line'], $this->error['end']),
+    );
+  }
+
+  public function getBlameCode(): ?string {
+    return $this->blame_code;
+  }
+
+  public function getPrettyBlame(): ?string {
+    return $this->getBlameCode();
+  }
+
+  <<__Memoize>>
+  public function getLintRule(): LintRule {
+    return new HHClientLintRule($this->error['code']);
+  }
+
+}
--- a/src/Linters/HHClientLinter.hack
+++ b/src/Linters/HHClientLinter.hack
@@ -1,0 +1,51 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+use namespace Facebook\TypeAssert;
+use namespace HH\Lib\{C, Vec};
+
+/**
+ * A linter as a proxy invoking `hh_client --lint`.
+ */
+final class HHClientLinter implements Linter {
+  use LinterTrait;
+
+  const type TConfig = shape();
+
+  const type TJSONResult = shape(
+    'errors' => vec<HHClientLintError::TJSONError>,
+    'version' => string,
+  );
+
+  public async function getLintErrorsAsync(
+  ): Awaitable<vec<HHClientLintError>> {
+    $lines = await __Private\execute_async(
+      'hh_client',
+      '--lint',
+      $this->getFile()->getPath(),
+      '--json',
+      '--from',
+      'hhast',
+    );
+    $hh_client_lint_result = TypeAssert\matches<this::TJSONResult>(
+      \json_decode(
+        C\firstx($lines),
+        /* assoc = */ true,
+        /* depth = */ 512,
+        \JSON_FB_HACK_ARRAYS,
+      ),
+    );
+    return Vec\map(
+      $hh_client_lint_result['errors'],
+      $error ==> new HHClientLintError($this->file, $error),
+    );
+  }
+}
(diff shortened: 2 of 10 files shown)


tests/HHClientLinterTest.hack
```diff
@@ -0,0 +1,56 @@
...
+final class HHClientLinterTest extends TestCase {
+  use LinterTestTrait;
+
+  /**
+   * The temporary directory to include testing source files to lint.
+   *
+   * Note that the temporary directory must be under the current directory,
+   * otherwise hh_client will not work.
+   */
+  const TMP_DIR = './.HHClientLinterTest.tmp';
```
fredemmott, line 22: `__DIR__ . '/../.var/tmp/hhast/HHClientLinterTest'` for consistency with other temp files used by hhast, and so that they're already in .gitignore

src/Linters/HHClientLintError.hack
```diff
@@ -0,0 +1,60 @@
...
+
+enum ArcanistLintSeverityEnum: string as string {
+  ADVICE = 'advice';
+  AUTOFIX = 'autofix';
+  WARNING = 'warning';
+  ERROR = 'error';
+  DISABLED = 'disabled';
+}
+
+class HHClientLintError extends LintError {
```
fredemmott, line 20: final class

src/Linters/HHClientLinter.hack
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Facebook\HHAST;
+
+use type Facebook\HHAST\{BaseLinter};
+use namespace Facebook\TypeAssert;
+use namespace HH\Lib\{C, Vec};
+
+final class HHClientLinter extends BaseLinter {
+
+  const type TConfig = shape();
```
fredemmott, line 18: The active lints need to be configurable; perhaps:

```
shape(
  ?'ignore_except' => keyset<int>,
  ?'ignore' => keyset<int>,
 )
```

Both would default to empty; if something is present in both, that should be a runtime error.

Additionally, there should probably be a hardcoded list of lints to already ignore, e.g. 5583 - await in a loop - is known buggy, and has a pure Hack implementation in HHAST and Facebook www already. The current list is HackLintRule::HACK_LINT_IGNORED_CODES in FB WWW, mostly being known-bad linters that should be fixed at some point, but for 5583/AWAIT_IN_LOOP, this is permanently replaced by the linters written in Hack so blocking it here (and in FB) is a band aid - the right fix is probably to delete that one from the ocaml code.

fredemmott (changes requested):
- test failure: either use a lint for the test that exists in all supported versions, or increase the minimum version (grep for the old one, probably composer.json and .github/workflows/*)
- needs adding to `LintRunConfig::NON_DEFAULT_LINTERS` so it's include in `all` runs.
  - this will make it so that it's include in HHAST's own runs; probably best to exclude this linter explicitly in hhast-lint.json for now, then fix them up separately

src/Linters/HHClientLintError.hack
```diff
@@ -0,0 +1,60 @@
...
+  public function getRange(): ((int, int), (int, int)) {
+    return tuple(
+      tuple($this->error['line'], $this->error['start']),
+      tuple($this->error['line'], $this->error['end']),
+    );
+  }
+
+  <<__Override>>
+  public function getBlameCode(): string {
+    return (string)$this->error['code'];
```
fredemmott, line 57: This should be returning the hack source code that is responsible for the error, not the numeric error code.

It might be reasonable to change `LintError::getBlameCode`'s default implementation to use getRange() and read it `getFile()`, instead of requiring it to be manually implemented here.

That said, the numeric lint code does need to be exposed. HHAST's current linters are all one-linter-per-class, so it doesn't vary by lint error. This needs to change.

These needs to be reflected in:
- HHAST_IGNORE_ERROR[code] must work
- lint errors in the CLI
- lint errors in vscode

Searching for `get_class` and `getLinterName` is likely a good way to find both these places, and any others that are appropriate

fredemmott (changes requested):

src/Linters/HHClientLintError.hack
Atry: Do you mean `HH_IGNORE_ERROR[code]`? I am not feeling comfortable sharing the same marker format with hhast linters.

fredemmott: Nope, `HHAST_IGNORE_ERROR[code]`; `HHAST_IGNORE_ERROR[LinterName]` is the existing suppression method used throughout.

Definitely don't want to re-use HH_ - they'd be removed when running `hh_client --remove-dead-fixmes`.

Atry: But why? Suppose there is an open source library synced from internal repository, we would like the marker be understood by both internal CI and the hhast linter, right?

fredemmott: In the past, that's not been a major concern due to there being no shared linters; with this, it's probably worth revisiting #351, but whatever's done should be consistent for all linters available via hhast - people should be able to use the same syntax regardless of if a linter is implemented in Hack or OCaml

Atry: Fair enough! I agree https://github.com/hhvm/hhast/pull/351 should be a good direction. But I would not address other marker format in this PR, instead `HHAST_*` markers should be a good start in this PR.

src/Linters/HHClientLinter.hack
Atry: I will implement the config in another PR as this PR is already too large

src/Linters/HHClientLintError.hack
Atry: See https://github.com/hhvm/hhast/pull/388 for `HHAST_IGNORE_ERROR[code]`

Atry: > lint errors in the CLI
> lint errors in vscode

Will be implemented in next couple of PRs

Atry:
> * test failure: either use a lint for the test that exists in all supported versions, or increase the minimum version (grep for the old one, probably composer.json and .github/workflows/*)

Fixed

> * needs adding to `LintRunConfig::NON_DEFAULT_LINTERS` so it's include in `all` runs.
>   
>   * this will make it so that it's include in HHAST's own runs; probably best to exclude this linter explicitly in hhast-lint.json for now, then fix them up separately

Will be addressed in another PR

src/Linters/HHClientLintError.hack
```diff
@@ -0,0 +1,73 @@
...
+
+  public function getBlameCode(): ?string {
+    return $this->blame_code;
+  }
+
+  public function getPrettyBlame(): ?string {
+    return $this->getBlameCode();
+  }
+
+  public function getLintRule(): LintRule {
```
fredemmott, line 69: this should probably be memoized or passed in to the constructor given that it's never going to change for a particular instance

src/Linters/HHClientLintError.hack
```diff
@@ -0,0 +1,73 @@
...
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+enum ArcanistLintSeverityEnum: string as string {
```
fredemmott, line 12: remove 'Enum' from name

fredemmott (approved):
note followup PRs (to add to non-default linters, and the configurability) are required before this can actually be used outside of unit tests.

src/Linters/HHClientLintError.hack
Atry: By the way, Hotspot JVM is good at optimizing allocation for short-lived objects like this. Is HHVM able to eliminate the allocation like JVM does?

fredemmott: Sorry, I'm not familiar enough with either VM:)

PR #369, opened 2021-11-04
Add the configuration for a dev container

Atry:
This PR adds the configuration for a dev container, which could be used in the [Visual Studio Code Remote - Containers](https://code.visualstudio.com/docs/remote/containers) extension.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,32 @@
+// For format details, see https://aka.ms/devcontainer.json.
+{
+  "name": "Hack",
+  "runArgs": [
+    "--init"
+  ],
+  "image": "hhvm/hhvm:latest",
+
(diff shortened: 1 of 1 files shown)


fredemmott:
Heh, thanks, I was looking at this yesterday (https://github.com/fredemmott/hhast/blob/codespaces/.devcontainer.json) for [GitHub Codespaces](https://github.com/features/codespaces) - though mine's untested as I wasn't aware vscode natively supported them.

Any preference or advantages/disadvantages for using a Dockerfile rather than using the original image with postCreateCommand for composer?

Atry:
> Any preference or advantages/disadvantages for using a Dockerfile rather than using the original image with postCreateCommand for composer?

Good idea! I just amended the commit as you suggested.

Atry:
截屏2021-11-04 上午9 15 52

PR #370, opened 2021-11-04
Allow the devcontainer to use the local ssh keys

Atry:
This commit would allow for `git push` from the devcontainer

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,7 +2,9 @@
 {
   "name": "Hack",
   "runArgs": [
-    "--init"
+    "--init",
+    "--volume",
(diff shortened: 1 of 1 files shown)


fredemmott:
I think this should be doable with the `mounts` argument instead, which will be ignored by github codespaces etc

fredemmott:
This also sounds like it's meant to 'just work': https://code.visualstudio.com/docs/remote/containers#_using-ssh-keys

fredemmott (changes requested):
yep, this shouldn't be needed.

I let vscode set up the container for hhast, then in terminal in vscode:
 - made sure the terminal was inside the container
 - ran `ssh-add -l`
 - saw my keys

Atry:
I executed `ssh-add` and it just work. It seems that the mounting is unnecessary! Thank you!

PR #371, opened 2021-11-04
Add MemoizedLineLinterTrait

Atry:
This commit adds `MemoizedLineLinterTrait`, which memoizes lines of current source file, improving the performance of `getLine`.

### DIFF
--- a/src/Linters/LineLinter.hack
+++ b/src/Linters/LineLinter.hack
@@ -10,19 +10,10 @@
 namespace Facebook\HHAST;
 
 use type Facebook\HHAST\{BaseLinter};
-use namespace HH\Lib\{C, Str, Vec};
+use namespace HH\Lib\{C, Vec};
 
 abstract class LineLinter<+Terror as LineLintError> extends BaseLinter {
-
-  public function getLinesFromFile(): vec<string> {
-    $code = $this->getFile()->getContents();
-    $lines = Str\split($code, "\n");
-    return vec($lines);
-  }
-
-  public function getLine(int $l): ?string {
-    return idx($this->getLinesFromFile(), $l);
-  }
+  use MemoizedLineLinterTrait;
 
   <<__Override>>
   public async function getLintErrorsAsync(): Awaitable<vec<Terror>> {
@@ -45,13 +36,6 @@ abstract class LineLinter<+Terror as LineLintError> extends BaseLinter {
     }
 
     return $errs;
-  }
-
-  /** Check if this linter has been disabled by a comment on the previous line.
-   */
-  protected function isLinterSuppressed(string $previous_line): bool {
-    return Str\contains($previous_line, $this->getFixmeMarker()) ||
-      Str\contains($previous_line, $this->getIgnoreSingleErrorMarker());
   }
 
   /** Parse a single line of code and attempts to find lint errors */
--- a/src/Linters/MemoizedLineLinterTrait.hack
+++ b/src/Linters/MemoizedLineLinterTrait.hack
@@ -1,0 +1,35 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+use namespace HH\Lib\Str;
+
+trait MemoizedLineLinterTrait {
+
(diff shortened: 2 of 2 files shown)


Atry:
This trait is also used #368 for `HHClientLinter`

fredemmott (changes requested):
This shouldn't be memoized, as:
- it means that a copy will be retained in memory of all files that have lint errors for this linter
- if multiple line-based linters have an error in the file, multiple copies will be retained
- it will only have a benefit if the file contains multiple lint errors, which we should assume is a rare case

This could be slightly improved by moving this to be part of the `File` class so it is shared between linters, but given that raised lint errors should be fixed, so lint errors should be rare, the potential CPU savings don't seem worth the memory

Atry:
I spend a while to understand what you mean. The problem is the reference from `LintError`s to the `Linter`, which extends the life of the `Linter` and holds the memory of memoized lines.

Thanks for the feedback, I will find another approach.

Atry:
Anyway, I think this issue expose a pitfall in our Hack object model. As you might know, I wrote Haskell in the past two years, where the reference to the upstream data would be released automatically once a thunk gets resolved. In contrast, the reference to upstream data would be kept forever, even after all the memoized functions get calculated.

fredemmott:
> The problem is the reference from LintErrors to the Linter

Perhaps this should entirely go; it's based on an implicit assumption that each linter raises one kind of error (i.e. all errors from the same error should be reported/handled similarly), but with #371 that stops being true.

> model, releasing

The problem is that __Memoize has been (ab)used throughout several codebases as a guaranteed 'run only once', despite that being an implementation detail. We can't realistically change that.

However, once coeffects are enforced by the runtime, we will hopefully start to be able to  do 'smart' memoization on pure functions.

Atry:
> Perhaps this should entirely go;

I am not sure how feasible to break backward compatibility. The constructor of `LintError` is public, if we remove the parameter, all the third-party liners will be broken.

fredemmott:
We rarely break backwards compatibility, but that's more out of a lack of need than a policy. Doing #371 cleanly feels like a good enough reason.

Atry:
I tried to remove the linter from the error, then I find that there are some use cases to get the linter from the error, which seem valid use cases:

https://github.com/hhvm/hhast/search?q=%22error-%3EgetLinter%28%29%22

Atry:
No worries, I will figure out another approach doing this

fredemmott:
> based on an implicit assumption that each linter raises one kind of error

Those calls fetching the linter are based on this assumption: we're fetching/displaying the class of the linter as currently, linter class == error code.

While the class will remain useful debugging information, it needs to stop being used as an error code.

fredemmott:
In particular, `LintError` has no function to retrieve an error code, only the source code (and linter).

Atry:
The reason why it is hard to solve the problem in a backward compatible way is that we did not distinguish between the API for tools that run linters, and the SPI for lint rule authors.

Ideally we should provide:
1. an API, where we can make covariant changes without breaking the backward compatibility
2. a SPI, where we can make contravariant changes without breaking the backward compatibility

fredemmott:
This does not need solving in a backwards-compatible way; while we don't do unnecessary breakages, trying particularly hard for it does not really make sense for this project:

Almost all things that use HHAST as a library use the AST (well, CST) nodes, and these - and their relations - exactly match Hack's parser, which itself does not have backwards compatibility guarantees.

For example, https://github.com/hhvm/hhast/commit/8de2e6b423750587e4f6e8fd97a88e8bc2d44086 is a BC-incompatible change, but it's unavoidable while keeping HHAST's core goals.

Atry:
I agree it's not worth to support error codes in a fully backward compatible way, because it would be difficult and tricky. However, I do hope we could have a chance to separate API and SPI, easing the changes in the future without breaking backward compatibility.

fredemmott:
I don't think an SPI is useful for this project, as it *must* exclude codegen/syntax/, which removes most of the utility.

Atry:
Is it possible to separate linters and AST into multiple libraries?

fredemmott:
We could do, but while that would let us technically offer APIs/SPIs, I don't think it would help much in practice; it would also make parser updates much more painful as we'd have a circular dependency: core/asts would dev-depend on linters which would hard-depend on ASTs.

Atry:
I implemented `lintMarkers` to memoize marker and line numbers instead of this PR. See https://github.com/hhvm/hhast/pull/384

PR #372, opened 2021-11-04
Configure code as the editor for the git command line

Atry:


fredemmott (commented):
I'd rather avoid having user preferences committed to the repo; that said:

- I think you'd want `code --wait` instead
- As long as you want the same configuration inside and outside the container, `git config --global core.editor "code --wait"` outside the container should do the trick without needing to set it for every project - vscode implicitly copies your git configuration into the container.

Atry:
> I think you'd want `code --wait` instead
I wasn't aware of the option. Thank you!

> As long as you want the same configuration inside and outside the container, `git config --global core.editor "code --wait"` outside the container should do the trick without needing to set it for every project - vscode implicitly copies your git configuration into the container.

Outside container git config definitely also works, but it still make sense to set VSCode as the default editor for the devcontainer because:

- `hhvm/hhvm` does not include any editor, not even `vi`
- The devcontainer, on top of `hhvm/hhvm`, include only one editor, which is VSCode
- The developer might have been set `core.editor` as some another editor, which does not exist in the devcontainer

fredemmott:
That makes sense, but:
- I dislike that this forces a setting, rather than sets a default
- I feel that's an argument that vscode should do this by default; it doesn't belong in per-project configurations

Also, either way can be configured per-user via a dotfiles repo, with a setup.sh that checks for docker envs and vscode. It'd be reasonable to make a setup script install vim and set EDITOR, but this will then override it.

Atry:
There are scripts to set up `GIT_EDITOR` in the official dev containers. 
https://github.com/microsoft/vscode-dev-containers/search?q=GIT_EDITOR

Instead of directly using `hhvm/hhvm`, we might want to build a more sophisticated Hack dev container based on the common scripts in `microsoft/vscode-dev-containers` repository.

fredemmott:
Yep, in particular it might be reasonable for us to pull in https://github.com/microsoft/vscode-dev-containers/blob/6cbf5a949d62d031ecaf25006db043dbc6138704/script-library/common-debian.sh verbatim

PR #375, opened 2021-11-06
Separate SingleRuleLinter into Linter and LintRule

Atry:
Reviewing each commit one by one would be easier, because there are multiple commits, which are too messy when reviewing them together.

### DIFF
--- a/src/Linters/LintError.hack
+++ b/src/Linters/LintError.hack
@@ -1,0 +1,36 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+/**
+ * A problem detected by a lint rule
+ *
+ * LintError is more generic than SingleRuleLintError because LintError is
+ * associated with any LintRule, not necessarily a SingleRuleLinter.
+ */
+<<__Sealed(
+  SingleRuleLintError::class,
+  // HHClientProblem::class
+)>>
+interface LintError {
+  public function getFile(): File;
+
+  public function getPosition(): ?(int, int);
+
+  public function getRange(): ?((int, int), ?(int, int));
+
+  public function getDescription(): string;
+
+  public function getBlameCode(): ?string;
+
+  public function getPrettyBlame(): ?string;
+
+  public function getLintRule(): LintRule;
+}
--- a/src/Linters/LintRule.hack
+++ b/src/Linters/LintRule.hack
@@ -19,8 +19,14 @@ namespace Facebook\HHAST;
 )>>
 interface LintRule {
   /**
-   * The name of the lint rule, which can be used to determine markers to
-   * suppress lint errors.
+   * The human readable name of the lint rule, which can be used to report
+   * the lint error.
    */
   public function getName(): string;
+
+  /**
+   * The error code of the lint rule, which can be used to determine markers to
+   * suppress lint errors.
+   */
+  public function getErrorCode(): string;
 }
--- a/src/Linters/LintRule.hack
+++ b/src/Linters/LintRule.hack
@@ -1,0 +1,26 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+/**
+ * An abstract lint rule that provides a single error reason, which could be
+ * either a HHAST linter or a lint rule written in OCaml.
+ */
+<<__Sealed(
+  SingleRuleLinter::class,
+  // HHClientLintRule::class,
+)>>
+interface LintRule {
+  /**
+   * The name of the lint rule, which can be used to determine markers to
+   * suppress lint errors.
+   */
+  public function getName(): string;
+}
--- a/src/Linters/Linter.hack
+++ b/src/Linters/Linter.hack
@@ -1,0 +1,38 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+/**
+ * The process to find problems against a single source file.
+ *
+ * Problems found by a Linter could associated with different LintRules,
+ * especially when the Linter is a proxy calling other linting tools.
+ */
+<<
+  __Sealed(
+    SingleRuleLinter::class,
+    // HHClientLinter::class
+  )
+>>
+interface Linter {
+  abstract const type TConfig;
+
+  public function getLintErrorsAsync(): Awaitable<vec<LintError>>;
+
+  public static function shouldLintFile(File $_): bool;
+
+  public static function newInstance(File $file, ?this::TConfig $config): this;
+
+  public function isLinterSuppressedForFile(): bool;
+
+  public function getFile(): File;
+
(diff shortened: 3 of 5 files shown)


src/Linters/BaseLinter.hack
```diff
@@ -13,13 +13,24 @@ use namespace Facebook\TypeAssert;
 use type Facebook\HHAST\File;
 use namespace HH\Lib\{C, Str};
 
+/**
+ * An HHAST linter
+ */
 <<__ConsistentConstruct>>
-abstract class BaseLinter {
+abstract class BaseLinter implements LintRule, Linter {
```
fredemmott, line 20: The naming is now misleading, as `BaseLinter` is no longer the base linter type. This should be changed, even though it's more of a BC break.

If I'm understanding the intent correctly, perhaps:

```Hack
abstract class SingleRuleLinter implements LintRule, Linter {
```

src/Linters/LintError.hack
```diff
@@ -11,7 +11,10 @@ namespace Facebook\HHAST;
 
 use type Facebook\HHAST\File;
 
-class LintError {
+/**
+ * A problem detected by an HHAST linter
```
fredemmott, line 15: by a BaseLinter (or SingleRuleLinter)

src/Linters/Problem.hack
```diff
@@ -0,0 +1,36 @@
...
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+/**
+ * A problem detected by an HHAST linter or an OCaml lint rule.
+ *
+ * Problem is more generic than LintError because Problem is associated with a
```
fredemmott, line 15: the naming here is also unclear; all problems are raised by a linter, so the distinction with LintError is pretty unintuitive. It's also a bit too generic - nothing in the name corresponds to the linting framework, as apposed to the other parts of HHAST.  For example, is it a migration 'problem'?

Perhaps call this interface 'LintError', and rename that class 'SingleRuleLinterError'?

fredemmott (changes requested):
The implementation/structure seem great, but the naming is very confusing. Changing it will be a BC break, but I think the clarity/tech-debt avoidance is worth it here.

src/Linters/Problem.hack
Atry: How about `LintProblem`? The terminology problem is consistent with VS Code and other IDEs. On the other hand,  error does not align with other libraries or applications, because `error` is usually considered as a severity level in hh_client and VS Code.

fredemmott: No strong feelings on what's more correct - though:
- if done, it should be consistent, i.e. all lint-related'*Error*' classes should be renamed
- if done consistently, it's a fairly big breaking change, and I'm not sure it's worth it
- if done, weak conflicting feelings:
  - the VSCode internal term is 'diagnostic'
  - IME most people call them 'lint errors', so it's my preference...
  - ... but, for now, in HHAST we report them all as WARNING severity diagnostics

src/Linters/LintError.hack
```diff
@@ -0,0 +1,36 @@
...
+ *
+ */
+
+namespace Facebook\HHAST;
+
+/**
+ * A problem detected by a lint rule
+ *
+ * LintError is more generic than SingleRuleLintError because LintError is associated with a
+ * LintRule, not a SingleRuleLinter.
```
fredemmott, line 16: > any LintRule, not necessarily a SingleRuleLinter

src/Linters/LintError.hack
```diff
@@ -0,0 +1,36 @@
...
+ */
+<<__Sealed(
+  SingleRuleLintError::class,
+  // HHClientProblem::class
+)>>
+interface LintError {
+  public function getPath(): string;
+
+  public function getPosition(): ?(int, int);
```
fredemmott, line 25: I don't feel strongly about this, but if it's going to be a BC-heavy release, maybe it's time to remove this for API simplification? it's equivalent for the first part of the getRange tuple, and the second part is nullable.

src/Linters/LintRule.hack
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * An abstract lint rule that provides a single error reason, which could be
+ * either a HHAST linter or a lint rule written in OCaml.
+ */
+<<__Sealed(
+  SingleRuleLinter::class,
+  // HHClientLintRule::class,
+)>>
+interface LintRule {
+  public function getName(): string;
```
fredemmott, line 21: - add a doc comment here explaining that it's used for suppression codes and showing for users
- perhaps add a sepaate 'getSuppressionCode', with default implementations in the base classes that return getName()

src/Linters/Linter.hack
```diff
@@ -0,0 +1,39 @@
...
+
+  public static function shouldLintFile(File $_): bool;
+
+  public function __construct(File $file, ?this::TConfig $config);
+
+  public function isLinterSuppressedForFile(): bool;
+
+  public function getFile(): File;
+
+  static public function typeAssertConfig(mixed $config): this::TConfig;
```
fredemmott, line 38: Can you experiment (in a different PR) with removing this?

Right now this uses the TypeAssert library, but that's because HHAST is older than the `is` and `as` operators, which should work for reasonable types of `this::TConfig`; we shouldn't have a method if `$value as this::TConfig` is usable.

search on github is also not showing me any callers to this?

fredemmott (approved):
Fine to merge as-is, either address before merging or a separate PR

src/Linters/LintError.hack
```diff
@@ -9,52 +9,28 @@
...
+ *
+ * LintError is more generic than SingleRuleLintError because LintError is associated with a
+ * LintRule, not a SingleRuleLinter.
+ */
+<<__Sealed(
+  SingleRuleLintError::class,
+  // HHClientProblem::class
+)>>
+interface LintError {
+  public function getPath(): string;
```
fredemmott, line 23: This needs to return a `File`, as linters can be ran on stuff that doesn't exist on disk (or doesn't match), especially when the linters are being ran interactively while editing the file in an IDE

src/Linters/LintRule.hack
Atry: Since the suppression logic is in the linter, we don't have to create `getSuppressionCode` here. Instead we can provide a trait for implementing Linters filtering out lint errors.

Anyway, we have chance to add more methods into `LintRule`. This PR aims minimize the API changes, especially when `LintRule` is sealed, allowing adding methods into it while keeping backward compatibility.

src/Linters/Linter.hack
```diff
@@ -0,0 +1,39 @@
...
+  __ConsistentConstruct,
+>>
+interface Linter {
+  abstract const type TConfig;
+
+  public function getProblemsAsync(): Awaitable<vec<LintError>>;
+
+  public static function shouldLintFile(File $_): bool;
+
+  public function __construct(File $file, ?this::TConfig $config);
```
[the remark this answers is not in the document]
Atry: Exactly. Hack does not support method overloading, making it impossible to deprecate constructors.

src/Linters/Linter.hack
Atry: Will do!

src/Linters/LintError.hack
Atry: Will do in another PR

src/Linters/LintRule.hack
fredemmott: > Since the suppression logic is in the linter

I think this makes sense for SingleRuleLinters, but as we want separate suppression codes per lint rule, they should probably be here, or separately in the lint error. The suppressio ncode for hhclientlinter will vary by concrete rule/error.

src/Linters/LintError.hack
Atry: https://github.com/hhvm/hhast/pull/385

src/Linters/LintRule.hack
Atry: `getErrorCode` added

src/Linters/Linter.hack
Atry: #386

PR #376, opened 2021-11-06
Reuse getLinterName()

Atry:


fredemmott (approved):
Please manually rebase before landing and make sure that the version change commit vanishes

PR #377, opened 2021-11-08
Upgrade HHVM version to 4.134

Atry:
The upcoming commits in #368 will include a test against an hh_client lint rule not in 4.123.


Atry:
I created another PR #378

PR #380, opened 2021-11-08
Migrate the internal implementation to use Linter instead of BaseLinter

Atry:


src/__Private/LSPImpl/CodeActionCommand.hack
```diff
@@ -39,7 +39,7 @@ final class CodeActionCommand extends LSPLib\CodeActionCommand {
             return null;
           }
 
-          $linter = $e->getLinter();
+          $linter = $e->getLintRule();
```
fredemmott, line 42: rename var

src/__Private/LSPImpl/CodeActionCommand.hack
```diff
@@ -39,7 +39,7 @@ final class CodeActionCommand extends LSPLib\CodeActionCommand {
...
           if (!$linter is AutoFixingLinter<_>) {
```
fredemmott, line 43: `hh_client --lint`'s output does have fields for replacement, so could also be autofixing. I don't know if any linters currently do this, but it's definitely a future possibility, so `AutoFixingLinter` should probably also be split up to be a property of lint errors (or maybe rules) rather than the linters themselves

Definitely a separate PR for that though

src/__Private/LintRun.hack
```diff
@@ -116,7 +116,7 @@ final class LintRun {
     }
 
     try {
-      $errors = await $linter->getLintErrorsAsync();
+      $errors = await $linter->getProblemsAsync();
```
fredemmott, line 119: I think I missed this somewhere in the stack: this should consistently be 'Errors', not 'Problems'

src/__Private/LintRunCLIEventHandler.hack
```diff
@@ -70,13 +70,13 @@ final class LintRunCLIEventHandler implements LintRunEventHandler {
           $error->getDescription(),
           $colors ? "\e[0m" : '',
           $colors ? "\e[90m" : '',
-          \get_class($error->getLinter()),
+          \get_class($error->getLintRule()),
```
fredemmott, line 73: - For hh_client --lint, the class name of the lint rule is not useful; I think you want $error->getLintRule()->getName() here?
- given this use case, I think we should split up 'human-readable lint rule name' from 'code'; for example, while  I think we want `HHAST_IGNORE_ERROR[1234]`, I don't think we want the CLI to say `Linter: 1234` - maybe `Linter: Hack Lint 1234`.

src/__Private/LintRunConfig.hack
```diff
@@ -84,7 +84,7 @@ final class LintRunConfig {
     HHAST\ConsistentLineEndingsLinter::class,
   ];
 
-  const vec<classname<SingleRuleLinter>> NON_DEFAULT_LINTERS = vec[
+  const vec<classname<Linter>> NON_DEFAULT_LINTERS = vec[
```
fredemmott, line 87: Doesn't need changing here, but when HHClientLinter is added, this mechanism won't be sufficient

fredemmott (approved):

src/__Private/LSPImpl/CodeActionCommand.hack
Atry: Will do! Thank you!

PR #384, opened 2021-11-09
Add the `File::errorCodesToSuppress` method, which scans a source file for all lint markers

Atry:
This dict returned by `lintMarkers` can be reused by all the linters, reducing repeated string search.

### DIFF
--- a/src/File.hack
+++ b/src/File.hack
@@ -9,7 +9,7 @@
 
 namespace Facebook\HHAST;
 
-use namespace HH\Lib\{Dict, Keyset, Regex, Str, Vec};
+use namespace HH\Lib\{C, Dict, Keyset, Regex, Str, Vec};
 
 final class File {
   private function __construct(
@@ -26,34 +26,60 @@ final class File {
     return Str\split($this->contents, "\n");
   }
 
+  /** 1-based line number */
   const type TLineNumber = int;
 
   /**
-   * Returns the 1-based line number of lint markers in the file.
+   * The lint error code to suppress, which could be either a linter name or a
+   * lint rule ID.
+   */
+  const type TErrorCode = string;
+
+  /**
+   * Returns the error codes to suppress according to lint markers in the file.
    */
   <<__Memoize>>
-  public function lintMarkersForLineBasedSuppression(
-  ): dict<string, keyset<self::TLineNumber>> {
-    return $this->getLines()
-      |> Vec\map_with_key(
+  public function errorCodesToSuppress(): shape(
+    'whole_file' => keyset<this::TErrorCode>,
+    'single_instance' => dict<this::TLineNumber, keyset<this::TErrorCode>>,
+  ) {
+    $all_matches = $this->getLines()
+      |> Vec\map(
         $$,
-        ($line_index, $line_content) ==>
-          Regex\every_match($line_content, re"/(?<marker_name>\w+)\[\w+\]/")
+        ($line_content) ==> Regex\every_match(
+          $line_content,
+          re"/(?<marker_name>\w+)\[(?<error_code>\w+)\]/",
+        )
           |> Vec\filter(
             $$,
             $match ==> LintMarkerName::isValid($match['marker_name']),
           )
-          |> Vec\map($$, $match ==> tuple($line_index + 1, $match[0])),
-      )
-      |> Vec\flatten($$)
-      |> Dict\group_by($$, $line_marker ==> $line_marker[/* marker */ 1])
-      |> Dict\map(
-        $$,
-        $line_markers ==> Keyset\map(
-          $line_markers,
-          $line_marker ==> $line_marker[/* line number */ 0],
-        ),
+          |> Vec\partition(
+            $$,
+            $match ==>
+              $match['marker_name'] === LintMarkerName::HHAST_IGNORE_ALL,
+          ),
       );
+    return shape(
+      'whole_file' => Vec\map(
+        $all_matches,
+        $line_matches ==>
+          // the partition 0: file level markers
+          $line_matches[0]
+          |> Vec\map($$, $match ==> $match['error_code']),
+      )
+        |> Keyset\flatten($$),
+      'single_instance' => Vec\map_with_key(
+        $all_matches,
+        ($line_index, $line_matches) ==>
+          // the partition 1: line level markers
+          $line_matches[1]
+          |> Keyset\map($$, $match ==> $match['error_code'])
+          |> C\is_empty($$) ? null : tuple($line_index + 1, $$),
+      )
+        |> Vec\filter_nulls($$)
+        |> Dict\from_entries($$),
+    );
   }
 
   public function getPath(): string {
--- a/src/Linters/SuppressibleTrait.hack
+++ b/src/Linters/SuppressibleTrait.hack
@@ -55,19 +55,18 @@ trait SuppressibleTrait {
    */
   public final function isSuppressedForFile(File $file): bool {
     return C\contains_key(
-      $file->lintMarkersForLineBasedSuppression(),
-      $this->getIgnoreAllMarker(),
+      $file->errorCodesToSuppress()['whole_file'],
+      $this->getErrorCode(),
     );
   }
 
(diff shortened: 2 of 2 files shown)


src/File.hack
```diff
@@ -22,6 +22,45 @@ final class File {
     return $this->isDirty;
   }
 
+  private function getLines(): vec<string> {
+    return Str\split($this->contents, "\n");
+  }
+
+  const type TLineNumber1Base = int;
```
fredemmott, line 29: Based not base

'base' by itself is a bit too confusing with number bases, e.g. base 16 == hex.

src/File.hack
```diff
@@ -22,6 +22,45 @@ final class File {
...
+
+  /**
+   * Returns the 1-base line number of lint markers in the file.
+   */
+  <<__Memoize>>
+  public function lintMarkers(): dict<string, keyset<self::TLineNumber1Base>> {
```
fredemmott, line 35: lintMarkersForLineBasedSuppression

Not all linters should respect these, especially AST-based linters (they should require that the suppression be attached to the specific AST node).

That's probably the best for hh-client based linters; given ASTs are cached, it shoudln't be too wasteful. The position can be transformed to a node via HHAST\find_node_at_position() - and the AST identifies which comments correspond to which nodes.

if memoized, this does need sharing between linters, but it's not great for separation of concerns :(

src/File.hack
```diff
@@ -22,6 +22,45 @@ final class File {
...
+        $$,
+        ($line_index, $line_content) ==>
+          Regex\every_match($line_content, re"/(?<marker_name>\w+)\[\w+\]/")
+          |> Vec\filter(
+            $$,
+            $match ==> LintMarkerName::isValid($match['marker_name']),
+          )
+          |> Vec\map(
+            $$,
+            $match ==> shape(
```
fredemmott, line 47: use a tuple instead so it's not a dict lookup

src/File.hack
```diff
@@ -22,6 +22,45 @@ final class File {
...
+              'line_number' => $line_index + 1,
+              'marker' => $match[0],
```
fredemmott, line 49: Should this really be returning the marker rather than just the code?

src/File.hack
fredemmott: `TLineNumber` feels sufficient though (comparing to 'line index')

fredemmott: if you strongly feel that some extra clarity is needed `TLineNumber1First` is probably the close to the clearest

src/File.hack
Atry: We actually need two indices: the marker name and the code. I previously let it return a `dict<LintMarkerName, dict<TErrorCode, keyset<TLineNumber>>>`, which is roughly equivalent to the current return type `dict<string, keyset<TLineNumber1>>`. I changed to the current return type because it is closer to the existing way to look up a marker.

fredemmott: Do we actually need the marker name, especially if we bring in #351? I'd expect us to just allow any form of suppresion for any code - e.g. both `HHAST_IGNORE_ERROR[1234]` and `@lint-ignore 1234`

Atry: I expect the marker name would be configurable in some extent, e.g. whether support `@lint-ignore`, `HH_FIXME`, `HHAST_FIXME`, etc. 

Including the the marker name would allow the ability of configuration.

src/File.hack
Atry: I am not sure if HHVM support something similar to ES6 `WeakMap` to implement a non-intrusive cache.

src/File.hack
fredemmott: - HH_* will definitely not be used
- I don't see a need for this to vary by linter; I can imagine some particular projects wanting to:
  - insist that all suppressions use the same format (in which case, that could be a linter, or hhast-wide configuration)
  - require 'these linters should use @lint-ignore, these should use <rule>' - but this seems unlikely
  
For both of the latter cases, the logic should likely be centralized rather than reproduced in each linter (or the base class)

src/File.hack
Atry: It's not worth to introduce some mechanism of Dependency Injection in this simple library. 
We might be able to implement the [Cake Pattern](https://www.baeldung.com/scala/cake-pattern) here, but Hack `trait`s are more restrictive than Scala `trait`s. 

Maybe there is no simple solution for separation of concerns here.

src/File.hack
Atry: Another reason why the marker name information is needed is that I use this function to cache `HHAST_IGNORE_ALL` as well, which is treated differently from `HHAST_FIXME` and `HHAST_IGNORE_ERROR `.

fredemmott: we should also support `@lint-ignore-all` in the future though; ultimately every linter should be:
- splitting the marker into marker name and error code
- figuring out if it's a -all marker
- acting appropriately

How about rename it to `getLintSuppressions`, and the return type changed to:

```hack
shape(
  'all' => keyset<CODE_AS_STRING>,
  'lines' => dict<int, keyset<CODE_AS_STRING>>
)
```

... though then the doc comment should emphasize that `lines` should only be used by line-based linters.

This (and ASTLinter) should share a common source of truth about which names are 'all' vs just-one-occurence

Atry: I just updated this PR to separate file-level and line-level markers

src/File.hack
```diff
@@ -26,34 +26,63 @@ final class File {
...
+  /**
+   * Returns the error codes to suppress according to lint markers in the file.
    */
   <<__Memoize>>
-  public function lintMarkersForLineBasedSuppression(
-  ): dict<string, keyset<self::TLineNumber>> {
-    return $this->getLines()
-      |> Vec\map_with_key(
+  public function errorCodesToSuppress(): shape(
+    'error_codes_in_file_lint_markers' => keyset<this::TErrorCode>,
```
fredemmott, line 43: maybe 'whole_file' and 'single_instance' as the keys? Seems both clearer, and less verbose/repetitive at callsites

There's no syntactic difference between 'this suppression applies to the next line' and 'this suppression applies to the next AST node'

src/File.hack
```diff
@@ -26,34 +26,63 @@ final class File {
...
+        if (!C\is_empty($line_matches)) {
+          $line_number = $line_index + 1;
+          $error_codes = () ==> {
+            foreach ($line_matches as $match) {
+              if ($match['marker_name'] !== LintMarkerName::HHAST_IGNORE_ALL) {
+                yield $match['error_code'];
+              }
+            }
+          };
+          yield $line_number => keyset($error_codes());
```
fredemmott, line 78: remove generators, change back to map/filter pipelines.

1) it's much more idiomatic in hack
2) generator performance has not been a focus of optimization for years, and it shows - they are extremely ineffecient in HHVM compared to vec/dicts and the functional approach.

src/Linters/SuppressibleTrait.hack
```diff
@@ -55,19 +55,18 @@ trait SuppressibleTrait {
...
+      $file->errorCodesToSuppress()['error_codes_in_file_lint_markers'],
+      $this->getErrorCode(),
     );
   }
 
   /** Check if this linter has been disabled by a comment on the previous line.
    */
   public final function isSuppressedForLine(File $file, int $previous_line_number): bool {
-    $line_marker = $file->lintMarkersForLineBasedSuppression();
+    $line_erro_codes =
```
fredemmott, line 66: error

fredemmott (approved):
Fine for these to be fixed in followups.

fredemmott:
Does need green test run though

src/File.hack
Atry: It would be verbose without the functions in the `fb` namespace.

fredemmott: Verbose implementation details is better than using generators. For example, https://github.com/hhvm/hhast/commit/50c3722b8a775a772302caacceae36b22505f3ad was a 2x speedup in 2017, and arrays have continued to be optimized - it's likely the problem is now worse.

fredemmott (approved):

PR #385, opened 2021-11-09
Remove LintError::getPosition to minimize the API

Atry:


src/__Private/LSPImpl/CodeActionCommand.hack
```diff
@@ -107,7 +107,7 @@ final class CodeActionCommand extends LSPLib\CodeActionCommand {
       if ($code !== ($diagnostic['code'] ?? null)) {
         continue;
       }
-      if ($error->getPosition() !== $pos) {
+      if ($error->getRange() |> $$ is null || $$[0] !== $pos) {
```
fredemmott, line 110: Neat trick :) though could you change to this for readability?

```hack
$error_position = $error->getRange()[0] ?? null;
if ($error_position !== pos) {
```

The addition of 'is null' appears to be a behavior change here too?

`$foo['bar']['baz'] ?? $x` will evaluate to $x if any of the intermediates in the expression are unset or null.

src/__Private/LintRunCLIEventHandler.hack
```diff
@@ -60,7 +60,7 @@ final class LintRunCLIEventHandler implements LintRunEventHandler {
       : null;
 
     foreach ($errors as $error) {
-      $position = $error->getPosition();
+      $range = $error->getRange();
```
fredemmott, line 63:
```hack
$position = $error->getRange()[0] ?? null;
```

... then the later code is more readable

fredemmott (changes requested):
Please change before merging unless you're building a stack on this.

fredemmott (approved):

PR #387, opened 2021-11-09
Add LinterTrait, which provide the common functions for SingleRuleLinter and HHClientLinter

Atry:


src/Linters/LinterTrait.hack
```diff
@@ -0,0 +1,100 @@
...
+ *  LICENSE file in the root directory of this source tree.
+ *
+ */
+
+namespace Facebook\HHAST;
+
+use type Facebook\HHAST\File;
+use namespace HH\Lib\{C, Str};
+
+<<__ConsistentConstruct>>
```
fredemmott, line 15: if we want more BC in the future, this should be removed, and a static factory method specified instead.

fredemmott (commented):
This feels like it should be an abstract class instead: every linter should use this trait and implement the interface.

src/Linters/LinterTrait.hack
Atry: I let the constructor be private, therefore no subclass is able to create another constructor and call `parent()`. The signature of the private constructor can be changed while keeping backward compatibility.

fredemmott:
Yeah :( I suspect we'll revisit this once traits are reworked and formalize. We might end up doing base classes and traits - base classes for devx, where the base classes always pull in the traits - but as multiple inheritance isn't supported, SingleRuleLinter would have to use at least one of the traits directly

Atry:
> Yeah :( I suspect we'll revisit this once traits are reworked and formalize. We might end up doing base classes and traits - base classes for devx, where the base classes always pull in the traits - but as multiple inheritance isn't supported, SingleRuleLinter would have to use at least one of the traits directly

I think the best design in Hack considering the restriction in object model is to minimize the interfaces and to eliminate the demand of sharing implementation. We can then move helper methods into namespaces as top-level free functions.

fredemmott:
> Each time we need new methods in the interfaces, we create a new interface containing the new methods instead of changing existing interfaces.

If there's a semantic separation, i agree with this, but not purely as a BC mechanism; large sets of related hierarchies really hurt developer experience. BC vs devx is a tradeoff.

PR #388, opened 2021-11-10
Suppress lint error according to error code level markers

Atry:


tests/examples/HHClientLinter/invalid_null_check.hack.expect
```diff
@@ -1,7 +1,7 @@
 [
     {
-        "blame": null,
-        "blame_pretty": null,
+        "blame": "cannot_be_null is null",
+        "blame_pretty": "cannot_be_null is null",
         "description": "Invalid null check: This expression will always return `false`.\nA value of type `int` can never be null."
     }
 ]
```
fredemmott, line 7: (separate PR) It's probably worth adding range information to LinterTestTrait::getErrorAsShape() so that these tests also make sure the the information provided to IDEs is accurate. This will require updating many of the existing tests.

fredemmott (approved):

PR #395, opened 2021-11-17
Run tests on Github Actions

Atry:


.github/workflows/build-and-test.yml
```diff
@@ -26,5 +26,7 @@ jobs:
           skip_lint: true
       - name: Lint
         run: bin/hhast-lint
+      - name: Test
+        run: vendor/bin/hacktest tests/
```
fredemmott, line 30: This is already part of `hhvm/actions/hack-lint-test` above

fredemmott (changes requested):

.github/workflows/build-and-test.yml
Atry: I thought the action means "tests of hack lint".

PR #397, opened 2021-11-23
Use `vec` instead of `keyset` in `HHClientLinter::TConfig`

Atry:
`keyset` is not supported according to this comment https://github.com/hhvm/hhast/commit/d12431fae2b37c60adff94fce7407e9ea933007f#diff-ab40415ae577f35a66424f1507ecc7dc1dd41322b54f8a5095ccf1d3f8ceb3c6R72-R73


fredemmott (approved):
yep; we require that the deserialized json matches the provided shape, and keysets aren't representable directly in json.

PR #398, opened 2021-11-23
Dogfood HHClientLinter for hhast source code

Atry:
All the existing errors are ignored for now

### DIFF
--- a/hhast-lint.json
+++ b/hhast-lint.json
@@ -22,5 +22,12 @@
         "Facebook\\HHAST\\DataProviderTypesLinter"
       ]
     }
-  ]
+  ],
+  "linterConfigs": {
+    "Facebook\\HHAST\\HHClientLinter": {
+      "Comments (not an actual setting)":
+        "The following error codes are ignored with no good reason. They are just existing issues and need to be fixed.",
+      "ignore": [5562, 5583, 5568, 5607, 5609, 5611, 5616, 5618, 5624, 5628, 5630, 5639]
+    }
+  }
 }
--- a/src/__Private/LintRunConfig.hack
+++ b/src/__Private/LintRunConfig.hack
@@ -94,6 +94,7 @@ final class LintRunConfig {
     HHAST\NoFinalMethodInFinalClassLinter::class,
     HHAST\NamespacePrivateLinter::class,
     HHAST\DontHaveTwoEmptyLinesInARowLinter::class,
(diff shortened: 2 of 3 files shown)


src/__Private/LintRun.hack
```diff
@@ -151,10 +151,11 @@ final class LintRun {
...
       }
     }
-    $results = await Vec\map_async(
-      $files,
-      async $file ==> await $this->lintFileAsync($config, $file),
-    );
+    $results = vec[];
+    foreach ($files as $file) {
+      /* HHAST_IGNORE_ERROR[DontAwaitInALoop] Intentionally run in serial in case of out-of-memory */
+      $results[] = await $this->lintFileAsync($config, $file);
```
lexidor, line 157: Is this equally fast as the parallel version? If it is not, would an Async\Semaphore be a middle ground? I'd like to be able to configure the parallelism that I am comfortable with in the linter config. map_with_limited_concurrency_async would have less overhead than a semaphore, but you'd have to write that logic inline.

Atry: Updated

hhast-lint.json
```diff
@@ -22,5 +22,10 @@
         "Facebook\\HHAST\\DataProviderTypesLinter"
       ]
     }
-  ]
+  ],
+  "linterConfigs": {
+    "Facebook\\HHAST\\HHClientLinter": {
+      "ignore": [5562, 5583, 5568, 5607, 5609, 5611, 5616, 5618, 5624, 5628, 5630, 5639]
```
fredemmott, line 28: perhaps this should be changed again to `dict<int, string>`, with the string as ignored comments, to make the config more maintainable?

e.g. 

```
"ignore": [5562 => "linter description: ignore reason" ]
```


Relatedly, what are these, and why are they ignored here, rather than automatically for all, or being fixed here?

fredemmott (approved):

hhast-lint.json
Atry: Probably we don't want to use "vanilla" JSON setting files because comments are supported.

Atry: Added some an unused field as comment

Atry:
Limit to concurrency for `hh_client` to 2, which means 2*CPU cores of processes, otherwise it would still run out of memory in my Docker Desktop.

PR #399, opened 2021-11-23
Limit the maximum number of linter instances to run in parallel

Atry:
Note that #398 requires this PR, otherwise `HHClientLinter` would be running out of memory.

### DIFF
--- a/src/__Private/LintRun.hack
+++ b/src/__Private/LintRun.hack
@@ -9,8 +9,9 @@
 
 namespace Facebook\HHAST\__Private;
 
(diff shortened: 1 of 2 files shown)


fredemmott:
Is the number of linters the problem, or the number of `hh_client` subprocesses? You might want to look at generalizing/reuing `ParserQueue` just for the `hh_client --lint` subprocess calls, rather than the linters as a whole.

Also, what's running out of memory? The container/environment/system, or is HHVM hitting an internal limit?

Atry:
Close this PR in favor of #400 according to the feedback

PR #400, opened 2021-11-23
Limit the maximum number of hh_client processes to run in parallel

Atry:
`ParserQueue` is generalized and reused in this PR


fredemmott (approved):
Perhaps an enum instead?

PR #401, opened 2021-11-24
[ Do not merge] HHClientLinter does not spew lint errors

lexidor:
This PR runs hhast in GA. I removed the ignores for hh_client linter. I expected to have a failed CI result, but it passes. We know this code base contains awaits in loops and that hh_client linter reports an error for those. These awaits only have hhast DontAwaitInALoop suppression, no HHClient(Linter) suppression.

### DIFF
--- a/hhast-lint.json
+++ b/hhast-lint.json
@@ -27,7 +27,7 @@
     "Facebook\\HHAST\\HHClientLinter": {
       "Comments (not an actual setting)":
         "The following error codes are ignored with no good reason. They are just existing issues and need to be fixed.",
-      "ignore": [5562, 5583, 5568, 5607, 5609, 5611, 5616, 5618, 5624, 5628, 5630, 5639]
+      "ignore": []
     }
   }
 }


hhast-lint.json
```diff
@@ -27,7 +27,7 @@
     "Facebook\\HHAST\\HHClientLinter": {
       "Comments (not an actual setting)":
         "The following error codes are ignored with no good reason. They are just existing issues and need to be fixed.",
```
Atry, line 29: Also delete the comments?

Atry (commented):
Let me investigate why HHClientLinter did not fail the test

hhast-lint.json
lexidor: This PR is not intended to be merged. This project shouldn't be "hh_client --lint" clean yet. Once the cause of the CI passing is found, this PR can be trashed. I made it to confirm my suspicion that it wasn't my local setup at fault.

Atry:
the comment accidentally disabled the whole HHClientLinter

lexidor:
So the code that validates the configuration does not throw an uncaught exception on invalid input? I can't find the code that does thus validation anymore. I put it in BaseLinter, but this class is refactored away.

lexidor:
closing in favor of #403

Atry:
I rebased your commit on top of #403, and now the CI failed as expected https://github.com/Atry/hhast/runs/4315515822

PR #402, opened 2021-11-24
Remove the comment, which accidentally disabled the whole HHClientLinter

Atry:


lexidor:
You are missing the ... in your config type. Configurable linters /should/ allow for unknown keys. This allows linters to gain configuration options in the future without breaking when this option is specified in an older build. If adding this comment is the reason the linter failed, no future options can be added in a backwards compatible manner.

lexidor:
https://github.com/hhvm/hhast/blob/6087e969d02b7420f385a7a73486009d4cc3f792/src/__Private/LintRunConfig.hack#L262

This exception should have been thrown all the way to `__EntryPoint` when the extra key was added in the config file.

Atry:
Close this PR in favor of #403

PR #403, opened 2021-11-24
Allow for unknown fields in the configuration of HHClientLinter

Atry:
This PR allows for unknown fields in the configuration of `HHClientLinter` so that:
- we can use the additional fields as comments
- we can keep the forward compatibility when adding new fields

#Thanks @lexidor for discovering the issue in https://github.com/hhvm/hhast/pull/402#issuecomment-978066789

### DIFF
--- a/src/Linters/HHClientLinter.hack
+++ b/src/Linters/HHClientLinter.hack
@@ -19,7 +19,7 @@ final class HHClientLinter implements Linter {
   use SuppressibleTrait;
 
   const type TConfig =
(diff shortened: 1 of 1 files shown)


lexidor (approved):
Yes, these three dots should allow a comment in the configuration. We should also look at the reason why a type validation failure did not cause an uncaught exception. Does you local env emit lint errors if you empty the ignore list for HHClientLinter?

Atry:
> Does you local env emit lint errors if you empty the ignore list for HHClientLinter?
Yes, I can see the lint errors, and I had put all the existing error code in the ignore config

PR #406, opened 2021-11-25
Fix lint errors emitted by HHClientLinter

lexidor:
Intentionally skipped three, see hhast-lint.json

### DIFF
--- a/src/__Private/codegen/CodegenBase.hack
+++ b/src/__Private/codegen/CodegenBase.hack
@@ -19,9 +19,9 @@ use namespace HH\Lib\{C, Dict, Keyset, Str, Vec};
 
 abstract class CodegenBase {
   // Note to future maintainers:
-  // This object is cloned use `clone` in `->withoutHackfmt()`.
+  // This object is cloned using `clone` in `->withoutHackfmt()`.
   // If you add a reference (object) typed property, you must ensure
-  // proper cloning sematics manually.
+  // proper cloning semantics manually.
   private Schema\TSchema $schema;
   private bool $useHackfmt = true;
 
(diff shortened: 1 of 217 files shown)


src/__Private/codegen/CodegenBase.hack
```diff
@@ -38,10 +43,13 @@ abstract class CodegenBase {
...
 
-  private bool $useHackfmt = true;
-
   final public function withoutHackfmt(): this {
     invariant($this->useHackfmt, "can't disable hackfmt twice");
+    /* Use of the native `clone` operator.
+       This will become problematic when reference (object) typed properties are added.
+       All properties are primitive, structural, or value arrays as of November 2021.
+       Calling the constructor explcitly would cost a lot of needless re-`Vec\sort_by()` work.
+       HHAST_FIXME[5562] */
```
Atry, line 52: Shall we let `schema` be a memoized function to avoid recalculation of `Vec\sort_by`

Atry (approved):

fredemmott:
This is going to need cleanup once #408 is resolved

src/__Private/codegen/CodegenBase.hack
fredemmott: Schema's rather large, and memoize is wasteful

The more common way to address this is to make the constructor protected, and move the expensive computation that's currently in the constructor to a public static method

Atry: Currently `withoutHackfmt` is called immediately after a `CodegenBase` is created, therefore the constructor will be called twice if we remove the `clone` operator. Memoization would not cost more memory than the current implementation because anyway `schema` would be stored on `CodegenBase`. The problem the that memoization could solve is to avoid the `schema` calculation in the first call to the constructor, because the fist instance can be discarded before the memoized function getting executed.

PR #412, opened 2021-11-26
Mismatched configuration for a linter should fail the lint run (fix #405)

### LINKED ISSUE #405: [ Regression ] Mismatched configuration for a linter should fail the lint run
Repro:

Configure a linterConfig with an incorrect type (pass a string where an int is expected or alike). Run hhast-lint. Observe the lint run completing as normal. Expected result, exception bubbles up to the EntryPoint:

https://github.com/hhvm/hhast/blob/6087e969d02b7420f385a7a73486009d4cc3f792/src/__Private/LintRunConfig.hack#L271

Atry:
I added a `cwd` parameter to `LinterCLI` for the test purpose so we could test it without actual `chdir`, in case of interfering other tests.

### DIFF
--- a/src/__Private/LinterCLI.hack
+++ b/src/__Private/LinterCLI.hack
@@ -121,7 +121,7 @@ final class LinterCLI extends CLIWithArguments {
       if (C\is_empty($roots)) {
         await $err->writeAllAsync(
           'You must either specify PATH arguments, or provide a configuration'.
-          "file.\n",
+          "file, in which the `roots` setting is not empty.\n",
         );
         return 1;
       }
@@ -133,10 +133,9 @@ final class LinterCLI extends CLIWithArguments {
           if (\file_exists($config_file)) {
             /* HHAST_IGNORE_ERROR[DontAwaitInALoop] HHAST_IGNORE_ERROR[5583]*/
             await $err->writeAllAsync(
-              'Warning: PATH arguments contain a hhast-lint.json, '.
-              'which modifies the linters used and customizes behavior. '.
-              "Consider 'cd ".
-              $root.
+              'Warning: PATH arguments contain a lint config file at '.$path.
+              'hhast-lint.json, which modifies the linters used and '.
+              "customizes behavior. Consider 'cd ".$root.
               "; vendor/bin/hhast-lint'\n\n",
             );
           }
--- a/src/__Private/LinterCLI.hack
+++ b/src/__Private/LinterCLI.hack
@@ -19,6 +19,7 @@ final class LinterCLI extends CLIWithArguments {
   private bool $xhprof = false;
   private ?string $xhprofDotFile = null;
   private LinterCLIMode $mode = LinterCLIMode::PLAIN;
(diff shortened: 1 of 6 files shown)


Atry:
I don't understand why this PR fails on CI, because the tests look good on my dev container.
```
$ vendor/bin/hacktest tests/LinterConfigTest.hack
```
```
......

Summary: 6 test(s), 6 passed, 0 failed, 0 skipped, 0 error(s).
```

lexidor (approved):
Oh wow a missing throw, I am going to make a linter for this. I believe this already exists for Javascript. _Don't use new for side effects_

src/__Private/LinterCLI.hack
```diff
@@ -102,12 +111,12 @@ final class LinterCLI extends CLIWithArguments {
     $roots = $this->getArguments();
 
     if (C\is_empty($roots)) {
-      $config = LintRunConfig::getForPath(\getcwd());
+      $config = LintRunConfig::getForPath($this->cwd ?? \getcwd());
```
fredemmott, line 114: Why do you need this rather than specifying a root?

Atry: To keep the backward compatibility when `$cwd` is omitted

Atry: There is a warning if I specify the root

> Warning: PATH arguments contain a hhast-lint.json

fredemmott (changes requested):
- It shouldn't be called cwd, as it's very decoupled from the working directory. Given all it's used for is specifying the path to the config file, it'd be clearer to take an optional full path to the configuration file (using `LintRunConfig::getFromConfigFile()`)
- I don't see a need for this to be a special-cased thing, especially in a CLI class that can already have optionals; what do you think about adding a `--config-file=` argument instead, and making the whole thing more flexible?
- Let's also change that warning to include ", but $ABSOLUTE_PATH_TO_ACTIVE_CONFIG will be used"

Atry:
The `--config-file` flag is added

tests/LinterConfigTest.hack
```diff
@@ -87,6 +93,29 @@ final class LinterConfigTest extends HackTest {
...
+      'A linter threw an exception:',
+    );
+    expect($stderr->getBuffer())->toContainSubstring(
+      'Linter: Facebook\HHAST\Tests\InvalidConfigForLinter',
+    );
+    expect($stderr->getBuffer())->toContainSubstring(
+      'Linter: Facebook\HHAST\Tests\InvalidConfigForLinter',
+    );
+    expect($stderr->getBuffer())->toMatchRegExp(
+      re'/File\\: .*\\/test-data\\/unparsable_php_file\\.php/',
```
fredemmott, line 110: When regexes contain paths (or literal forward slashes more generally), improve readability by using a different delimiter, e.g.:

```hack
re#File\\: .*/test-data/unparseable_php_file\\.php#
```

fredemmott (approved):

Issue #409, opened 2021-11-26
Delete HHAST linters that duplicate good HHClientLinters


fredemmott:
Relates to #407 and #408 : we should try to avoid duplicating work, and also avoid raising two errors for the same semantic issue; it should be relatively rare to need to suppress two linters in the same line.

If the HHClientLinters are known bad, they should be automatically suppressed (and deleted from hh_client if they wont' be fixed).

If they're good, the HHAST linter should be deleted.

lexidor:
We are not affected by the BC break that this brings, but I bring it up anyway. HHClientLinter is a multi linter. If a sub rule was previously disabled for a certain path, devs would either have to:
 - live without all HHClientLinter lints in that path
 - live without that lint type in their code base
 - supp all issues of the now active linter in that path

Something like a linterConfig setting like this would work.
```JSON
"doNotRaiseErrorsFor": {"5562": [<path regexes>], "5633": [...]}
```

fredemmott:
It needs to be user-configurable before it's enabled in a release, but:
- that should ideally be via the existing 'disabledLinters' configuration option for consistency
- we shouldn't maintain code that duplicates built-in functionality
- we should provide good defaults even with 'all' setting.

Atry:
@lexidor 
We can let `overrides` in  `hhast-lint.json` also support `linterConfigs`

Atry:
Some users might want to not run `HHClientLinters` and some other linters in dependencies, because obviously you cannot expect all the third-party libraries updated their code base immediately after a new compiler version released.

fredemmott:
FWIW, if people want some level of upgrade safety, they should not use `all` or `default`, and should explicitly list all linters (and hhclient lint codes) that they want enabled for their project.

This will still raise errors when bugs are fixed in linters that lead to them raising more errors.

Atry:
Fixed in #427

PR #414, opened 2021-11-26
Support file specific linter config

Atry:
The file specific linter config is used in #417

### DIFF
--- a/src/__Private/LintRunConfig.hack
+++ b/src/__Private/LintRunConfig.hack
@@ -164,7 +164,7 @@ final class LintRunConfig {
     );
   }
 
-  public function getConfigForFile(string $file_path): self::TFileConfig {
+  private function relativeFilePath(string $file_path): ?string {
     $roots = Vec\map(
       $this->configFile['roots'],
       $s ==> Str\strip_suffix($s, '/').'/' |> Str\strip_prefix($$, './'),
@@ -175,6 +175,14 @@ final class LintRunConfig {
       $roots !== vec[] &&
       !C\any($roots, $root ==> Str\starts_with($file_path, $root))
     ) {
+      return null;
+    }
+    return $file_path;
+  }
+
+  public function getConfigForFile(string $file_path): self::TFileConfig {
+    $file_path = $this->relativeFilePath($file_path);
+    if ($file_path is null) {
       return shape(
         'linters' => keyset[],
         'autoFixBlacklist' => keyset[],
--- a/src/__Private/LintRunConfig.hack
+++ b/src/__Private/LintRunConfig.hack
@@ -14,6 +14,18 @@ use namespace HH\Lib\{C, Keyset, Str, Vec};
 use type Facebook\HHAST\Linter;
 
 final class LintRunConfig {
+
+  const type TOverride = shape(
+    // Which files this override applies to (uses `fnmatch()`)
+    'patterns' => vec<string>,
+    ?'builtinLinters' => NamedLinterGroup,
+    ?'extraLinters' => vec<string>,
+    ?'disabledLinters' => vec<string>,
+    ?'disabledAutoFixes' => vec<string>,
+    ?'disableAllAutoFixes' => bool,
+    ?'disableAllLinters' => bool,
+    ?'linterConfigs' => dict<string, dynamic>,
+  );
   const type TConfigFile = shape(
     // Where to lint, eg '[ "src/", "codegen/", "tests/" ]
     'roots' => vec<string>,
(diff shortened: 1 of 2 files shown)


tests/LinterConfigTest.hack
```diff
@@ -87,6 +93,29 @@ final class LinterConfigTest extends HackTest {
     )->toThrow(\Exception::class, 'is not of the correct type');
   }
 
+  public async function testInvalidConfigForCLI(): Awaitable<void> {
+    list($cli, $_, $stdout, $stderr) =
+      $this->getCLIInDirectory(vec[], self::TEST_DIRECTORY);
+    $exit_code = await $cli->mainAsync();
+    expect($stderr->getBuffer())->toMatchRegExp(
+      re'/A linter threw an exception\\:/',
+    );
```
lexidor, line 102: I see you use toMatchRegExp everywhere. If you use toContainSubString you can save yourself (and future readers) from having the reason about escaped regular expression characters.

fredemmott:
This is  probably a worthwhile change on itself, but I don't think it should be needed here: does `/*HHAST_IGNORE_ALL[1234]*/` not work for these? If not, that should be fixed:
- for consistency
- per-file suppressions via config is not scalable, especially as json's lack of support for trailing commas increases the risk of merge conflicts

lexidor:
> This is probably a worthwhile change on itself, but I don't think it should be needed here: does `/*HHAST_IGNORE_ALL[1234]*/` not work for these? If not, that should be fixed:

I believe this PR intends to address my comment on #409

This might also be a nice alternative to `.LICENSE_HEADER.hh.txt`. The LicenseHeaderLinter predates linter configuration and uses a well-known naming convention for setting per directory license headers instead. Once configuration is able to be declared in a override, this system can be supplemented using a configuration.

Atry:
> This is probably a worthwhile change on itself, but I don't think it should be needed here: does `/*HHAST_IGNORE_ALL[1234]*/` not work for these? If not, that should be fixed:
> 
> * for consistency
> * per-file suppressions via config is not scalable, especially as json's lack of support for trailing commas increases the risk of merge conflicts

I suppose people could use this PR to suppress linter issues in dependencies, where you cannot easily add `HHAST_IGNORE_ALL` comments.

fredemmott:
> I suppose people could use this PR to suppress linter errors in dependencies, where you cannot easily add HHAST_IGNORE_ALL comments.

People should not be linting their dependencies; lint-cleanness should be considered an implementation detail

fredemmott:
> This is probably a worthwhile change on itself,

Just to clarify what I meant here:

I think this PR is good, and should probably be merged

I don't think the rest of the stack should depend on it :)

Atry:
Rebased this PR to avoid dependencies between other PRs

hhast-lint.json
```diff
@@ -24,6 +24,14 @@
       "extraLinters": [
         "Facebook\\HHAST\\DataProviderTypesLinter"
       ]
+    },
+    {
+      "patterns": ["src/Migrations/*"],
+      "linterConfigs": {
+        "Facebook\\HHAST\\HHClientLinter": {
+          "ignore": [5624, 5639]
```
fredemmott, line 32: Filed #416 for removing this

src/Migrations/AddXHPChildrenDeclarationMethodMigration.hack
```diff
@@ -43,19 +43,14 @@ final class AddXHPChildrenDeclarationMethodMigration
...
-    if ($t->getLeading() !== $t->getLeadingWhitespace()) {
-      $ut = $use->getFirstTokenx();
-      return $in->withDeclarations(
-        $in->getDeclarations()->insertBefore(
-          $before,
-          $use->replace($ut, $ut->withLeading($t->getLeading())),
-        ),
-      )
-        ->replace($t, $t->withLeading(null));
-    }
```
fredemmott, line 55: Doesn't look like these changes are meant to be in this PR?

src/__Private/LintRunConfig.hack
```diff
@@ -39,21 +51,10 @@ final class LintRunConfig {
...
-        ?'disabledAutoFixes' => vec<string>,
-        ?'disableAllAutoFixes' => bool,
-        ?'disableAllLinters' => bool,
-      )
-    >,
+    ?'overrides' => vec<self::TOverride>,
     // Each linter may specify a type for itself.
     // The type of this key is effectively `dict<classname<T1 ... Tn>, Tx>`
-    ?'linterConfigs' => dict<string, dynamic>,
+    ?'linterConfigs' => dict<string, shape(...)>,
```
fredemmott, line 57: revert to dynamic, and use dynamic for the new option

https://github.com/hhvm/hhast/pull/364#discussion_r707531021

fredemmott (approved):
use dynamic instead of shape and remove XHP changes before merging

fredemmott:
XHP change seems needed because of this, but should be a separate PR (and separate commit in the main branch) as it's independent and should be reviewed (and potentially bisected or reverted) separately

src/Migrations/AddXHPChildrenDeclarationMethodMigration.hack
Atry: This change is to fix a 5607 linter error. I reviewed `$t->getLeading() !== $t->getLeadingWhitespace()` and I believe the condition is always true, therefore, I think the `if` clause is unnecessary.

Atry: Anyway, I will split the change to another PR

PR #417, opened 2021-11-30
Don't ignore 5607 lint error under src/Migrations and fix an existing 5607 lint error

Atry:
This PR is to fix a 5607 linter error under `src/Migrations`. I reviewed `$t->getLeading() !== $t->getLeadingWhitespace()` and I believe the condition is always `true`, therefore, I removed the `if` clause.

### DIFF
--- a/hhast-lint.json
+++ b/hhast-lint.json
@@ -24,6 +24,14 @@
       "extraLinters": [
         "Facebook\\HHAST\\DataProviderTypesLinter"
       ]
+    },
+    {
+      "patterns": ["src/Migrations/*"],
+      "linterConfigs": {
+        "Facebook\\HHAST\\HHClientLinter": {
+          "ignore": [5624, 5639]
+        }
+      }
     }
   ],
   "linterConfigs": {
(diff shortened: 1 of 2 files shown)


fredemmott:
can you unstack this (and leave out the configuration change for now), so that it can be tested/merged/reverted independently?

Atry:
Rebased

fredemmott (commented):
Can you split out the configuration change please? The goal is to have independently bisectable/revertible commits for different issues.

hhast-lint.json
```diff
@@ -24,6 +24,14 @@
       "extraLinters": [
         "Facebook\\HHAST\\DataProviderTypesLinter"
       ]
+    },
+    {
+      "patterns": ["src/Migrations/*"],
+      "linterConfigs": {
+        "Facebook\\HHAST\\HHClientLinter": {
+          "ignore": [5624, 5639]
```
fredemmott, line 32: *bump* - this is what I'm referring to

Atry: Currently the global ignore code list is `[5607, 5624, 5639]`, so this commit actually enabled 5607 error code in comparison to the global ignore list

PR #418, opened 2021-12-09
Suppress lint errors file by file

Atry:


codegen/syntax/AliasDeclaration.hack
```diff
@@ -1,11 +1,13 @@
...
  * This file is generated. Do not modify it manually!
  *
- * @generated SignedSource<<51b4ce4fe28d6124a26843b29b1f73f7>>
+ * @generated SignedSource<<d4fc81d224048c5ec9315137fe137818>>
  */
 namespace Facebook\HHAST;
 use namespace Facebook\TypeAssert;
 use namespace HH\Lib\Dict;
+/* HHAST_IGNORE_ALL[5607] 5607 is ignored because of false positives when comparing a generic to a typed value */
+/* HHAST_IGNORE_ALL[5624] HHAST_IGNORE_ALL[5639] 5624 and 5639 are ignored because they insist on using co(tra)variant generics. Could this break external consumers? */
```
fredemmott, line 10: This linter might also be good for the default blacklist - there's an ongoing discussion about removing it entirely at https://fb.workplace.com/groups/hackforhiphop/posts/7429940763721141

In short: It's accurate, but it's highly opinionated, and there's not consensus on the behavior it encourages being a good thing.

This isn't the same as the other codes we'd want to always-ignore, but "pretend it doesn't exist" may be the right thing to do for "may be deleted soon". Perhaps making it off-by-default but explicitly-includable may be better.

hhast-lint.json
```diff
@@ -24,23 +24,6 @@
...
-  "linterConfigs": {
-    "Facebook\\HHAST\\HHClientLinter": {
-      "Comments[5607] (not an actual setting)":
-        " - 5607 is ignored because of false positives when comparing a generic to a typed value",
-      "Comments[5624&5639] (not an actual setting)":
-        " - 5624 and 5639 are ignored because they insist on using co(tra)variant generics. Could this break external consumers?",
-      "ignore": [5607, 5624, 5639]
     }
-  }
+  ]
```
fredemmott, line 28: <3

fredemmott (approved):

hhast-lint.json
fredemmott: ... for the whole file changes, not specifically the bracket :)

Atry: Seems like I did something wrong when resolving conflicts

PR #419, opened 2021-12-09
Ignore 5583 error code from hh_client --lint

Atry:


src/Linters/HHClientLinter.hack
```diff
@@ -23,13 +23,21 @@ final class HHClientLinter implements Linter {
...
+
   <<__Memoize>>
   private function isErrorCodeConfiguredToIgnore(): (function(this::TErrorCode): bool) {
     $ignore_except = $this->config['ignore_except'] ?? null;
     $ignore = $this->config['ignore'] ?? null;
     if ($ignore is null) {
       if ($ignore_except is null) {
-        return $_ ==> false;
+        return $error_code ==> C\contains(self::DEFAULT_IGNORE, $error_code);
       }
```
lexidor, line 41: Should default ignore not always be honored? If I wish to ignore 5001, this enables everything ignored by default. This means I need to spell out every ignored linter again if I wish to ignore another one.

Atry: Good question. What if people want to disable DontAwaitInLoopLinter and not ignore 5583 `hh_client` lint rule if we always ignore 5583?

lexidor: I am just worried that people want to ignore a pedantic thing when introducing these new linters. Introducing all these lints in a big code base at once will lead to a gigantic PR. So we ignore sketchy null checks for now. These people are now getting served buggy results from known buggy linters.

Atry: > I am just worried that people want to ignore a pedantic thing when introducing these new linters. Introducing all these lints in a big code base at once will lead to a gigantic PR. So we ignore sketchy null checks for now. These people are now getting served buggy results from known buggy linters.

Why would the situation be different whether or not we allow users to not ignore those lint rules ignored by default?

fredemmott: > Good question. What if people want to disable DontAwaitInLoopLinter and not ignore 5583 hh_client lint rule if we always ignore 5583?

I think we should take this ignore list as 'pretend it doesn't exist', and not support this use case.

The correct fix for 5583 - and most of the others in the internal list - is to delete them entirely in the ocaml code. We should do that.

The purpose of copying the list to HHAST is to get us to a usable state quicker - we should still delete them from the ocaml code, but we need to be careful when doing that and it'll take a while to do that for all of them.

fredemmott (approved):

Issue #287, opened 2020-04-16
[ RFC ] What changes would you like to make to the hhast-lint.json format?


lexidor:
I'd like to rethink the hhast-lint.json config settings, since there are some short comings.
Here is just a short list of properties that I'd like to have, in no particular order.
This is not an issue that is meant to be implemented; it is meant to start a discussion.

*Updating HHAST should not cause more linters to be activated (causing failing travis builds).*

When upgrading hhast versions, linters that have been written for that version are automatically turned on. This makes upgrading hhast (which you are sometimes required to do, because of hhvm AST changes) a more painful process than it should be.

*There should be more granularity in the built-in linters than 'none' | 'default' | 'all'*

Currently `all`, doesn't mean all linters.
The only way to know that you are missing these linters is by digging into hhast and finding little gems like `NoEmptyStatementsLinter` or `FinalOrAbstractClassLinter`.
They might have been left out of the `all` option (NON_DEFAULT_LINTERS in the source) by mistake.

The `default` option currently excludes 8 linters, but it is not clear what makes a linter non-default.

The `none` option is made for people who require a stable output from their linters and who do not want and upgrade to introduce more linters that their code will fail under.
These people will list out every single linter they care about, one by one, and they will not be made aware of new and exiting linters.

*A hhast-lint.json file that is valid in version B should be valid for version A, even if linters that are declared in the hhast-lint.json file are not present.*

Currently, when you name a linter that does not exist, you get this error:
`Expected type 'Facebook\HHAST\BaseLinter', got type 'string'`
This stops HHAST dead in its tracks, since it is an uncaught exception.
I'd like for HHAST to report that linter X could not be found and continue linting with the rest of the linters.

*I would be nice to be able to refer to groups of linters, instead of individual linters.*

I'd like to be able to define `Lexidor's pedantic linter pack: 1.4`, `Fully autofixable linters: *`, `Code style linters: 0.8` and have that be recognized by hhast.

I would like to introduce a new hhast-lint.lock file.
This would allow us to:
 - Add new linters that would be enabled under the hhast-lint.json config, without having them pop-up upon upgrading hhast.
 - Break BC in the hhast-lint.json format, without bricking every travis job, since hhast-lint.lock would still work under the newer version of hhast.
 - Introduce a hhast-lint-update script, which will tell you which changes it is about to make.

jjergus:
> Updating HHAST should not cause more linters to be activated (causing failing travis builds).

[...]

PR #420, opened 2021-12-09
Add HHClientLinter to NON_DEFAULT_LINTERS

Atry:


fredemmott:
The change looks good, but I think we should run it on some projects - both ours, and others - first.

This is a lot of linters to enable in one go, and we should try to get some more signal on if this is currently a good configuration.

This also ties into #287 a bit: linter inclusion in general, 'all', and 'default' is not a good categorization system.

fredemmott:
What do you think about leaving this PR for now, instead:
- finish the other todos on https://github.com/hhvm/hhast/projects/1
- leaving it for an explicit opt-in
- ... but make a new release that supports the opt-in, announce on hhvm.com/blog/ , with details on configuration and asking for feedback?

PR #422, opened 2021-12-09
By default ignore 5639 error code from hh_client --lint

### LINKED ISSUE #421: Disable some lint rules by default
This linter might also be good for the default blacklist - there's an ongoing discussion about removing it entirely at https://fb.workplace.com/groups/hackforhiphop/posts/7429940763721141

In short: It's accurate, but it's highly opinionated, and there's not consensus on the behavior it encourages being a good thing.

This isn't the same as the other codes we'd want to always-ignore, but "pretend it doesn't exist" may be the right thing to do for "may be deleted soon". Perhaps making it off-by-default but explicitly-includable may be better.

_Originally posted by @fredemmott in https://github.com/hhvm/hhast/pull/418#discussion_r766003467_

Atry:
This PR fixes #421

### DIFF
--- a/src/Linters/HHClientLinter.hack
+++ b/src/Linters/HHClientLinter.hack
@@ -30,6 +30,14 @@ final class HHClientLinter implements Linter {
     5583 /* DontAwaitInALoop, which should have been covered by DontAwaitInALoopLinter */,
   ];
 
+  /**
+   * The error code that are ignored by default when both `ignore_except` and
+   * `ignore` are `null`
+   */
+  const keyset<int> DEFAULT_IGNORE_ERRORS = keyset[
+    5639 /* potential co(tra)variant marker, which is highly opinionated */,
+  ];
+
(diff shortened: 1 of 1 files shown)


src/Linters/HHClientLinter.hack
```diff
@@ -30,6 +30,14 @@ final class HHClientLinter implements Linter {
     5583 /* DontAwaitInALoop, which should have been covered by DontAwaitInALoopLinter */,
   ];
 
+  /**
+   * The error code that are ignored by default when both `ignore_except` and
+   * `ignore` are `null`
+   */
+  const keyset<int> DEFAULT_IGNORE_ERRORS = keyset[
+    5639 /* portential co(tra)variant marker, which is highly opinionated */,
```
fredemmott, line 38: po~~r~~tential co(*n*tra)variant marker

fredemmott (approved):

PR #431, opened 2022-01-07
DontCreateForwardingLambdasLinter

lexidor:
Before the addition of function references to Hack, you had to use `fun()` and `class_meth()` to create references to `$this`-less functions. This was hard to use when using namespaces, since the strings had to be fully qualified. Creating a forwarding lambda was syntactically more compact.
```HACK
namespace OurNamespace\AndTheSubNamespace;

Vec\map($v, fun('OurNamespace\\AndTheSubNamespace\\func'));
Vec\map($v, $x ==> func($x));
// Available for a while now
Vec\map($v, func<>);
```

We have many of these kinds of forwarding lambdas from those days. Function references are preferred, but I keep typing `$x ==> f($x)` out of habit / muscle memory. I hope this linter is useful for others.


Atry:
Forwarding lambdas might be useful for debugging, because it could added a layer of stack frame.

lexidor:
> Forwarding lambdas might be useful for debugging, because it could add a layer of stack frame.

This is indeed a good use for a forwarding lambda. This is useful when passing functions down (via return) or when storing a callable for later use. When passing a function as a callback to hsl-like functions, the call site is already included in the stack trace.

If a forwarding lambda was written for the purpose of improving the stack trace, I'd like to see an HH_IGNORE_ERROR that explains that this lambda is expressly added for that purpose.

```HACK
function pick_impl_for(It $it): TFuncType {
  if ($it->isSpecial()) { $impl = special_func<>; } else if ...
  /* HHAST_IGNORE_ERROR[DontCreateForwardingLambdas]
  We make a lambda here because we want to add this source location
  to the stack trace for debugging exceptions. */
  return $x ==> $impl($x);
}
```

Atry:
What [group](https://github.com/hhvm/hhast/discussions/428) would this lint belong to? It is definitely neither `fb-style` nor `errors`.

lexidor:
Hmmm, this linter is kinda hard to find a label for. The group `suggestions` is vague, but it doesn't fit in any of the others.

Atry (approved):
This is a decent PR with sophisticated examples, comments, and tests.

Looks good to me.

PR #391, opened 2021-11-11
Add HHClientLinterTestTrait for setting up the temporary source directory

Atry:

PR #433, opened 2022-01-20
Ignore temporary files from type checking

Atry:
It contains temporary files generated in #391

### DIFF
--- a/.hhconfig
+++ b/.hhconfig
@@ -1,4 +1,4 @@
-ignored_paths = [ "tests/examples/NonHackFileTest/php_files/*", "vendor/bin" ]
+ignored_paths = [ "tests/examples/NonHackFileTest/php_files/*", "vendor/bin", ".var/tmp/hhast" ]
 error_php_lambdas = true
 disable_xhp_children_declarations = false
 allowed_decl_fixme_codes=2053,4045,4047


.hhconfig
```diff
@@ -1,4 +1,4 @@
-ignored_paths = [ "tests/examples/NonHackFileTest/php_files/*", "vendor/bin" ]
+ignored_paths = [ "tests/examples/NonHackFileTest/php_files/*", "vendor/bin", ".var/tmp/hhast" ]
```
fredemmott, line 1: - What's in here? Does it have/need typecheckable file extensions?
- we can probably just ignore all of .var/

Atry: > What's in here? Does it have/need typecheckable file extensions?

It contains temporary files generated in [here](https://github.com/hhvm/hhast/pull/391/files#diff-cc0ff4a691608cd7f1c8866bf96abb52572353f214847a43730a4ce3bf7541acR27). They should not type check because the files might contain errors.

fredemmott: I *think* we can remove the .hack extension from them, so the typechecker won't look at them unless explicitly asked to - and hh_client --lint should trigger them being looked at:

```
fredemmott@fredemmott-fbmbp user-documentation % cat test
async function bar(): Awaitable<void> {
}
async function foo(): Awaitable<void> {
  while(true) {
    await bar();
  }
}
fredemmott@fredemmott-fbmbp user-documentation % hh_client --lint test
Lint[5583] Do not use await in a loop. It almost always incurs non-obvious serial fetching that is easy to miss. See https://fburl.com/awaitinloop for more information. [1]

test:5:5
    3 | async function foo(): Awaitable<void> {
    4 |   while(true) {
[1] 5 |     await bar();
    6 |   }
    7 | }
```

This gets us closer to a zero-config good state.


We should then explicitly ignore .php files as:
- they're deprecated
- the extension is required for .php files to be parsed properly by hack; any unrecognized extension (or extensionless file) is considered to be a hack file.

Atry: How about `.hack.in` extension? It would be more readable than no extension.

fredemmott: Sure, that works - though AIUI these should not be user visible?

Atry: I think the extension does effect the type checker behavior:

https://github.com/hhvm/hhast/runs/4901949350?check_suite_focus=true

`hh_client --lint` would not report errors any more if the extension is `.hack.in`.

fredemmott: the extension affects the automatic "do I care about this file?" stuff when walking the tree or looking for modifications.

*if* the typechecker is looking at a file - either because yes, it cares about that file by default, or if it's been explicitly asked to look at it - only `.php` is special, all other files are treated as `.hack`

> `hh_client --lint` would not report errors any more if the extension is `.hack.in`.


Based on the above, this shouldn't be the case, and doesn't appear to be:

```
cat test.hack.in
async function bar(): Awaitable<void> {
}
async function foo(): Awaitable<void> {
  while(true) {
    await bar();
  }
}
$ hh_client --lint test.hack.in
Lint[5583] Do not use await in a loop. It almost always incurs non-obvious serial fetching that is easy to miss. See https://fburl.com/awaitinloop for more information. [1]

test.hack.in:5:5
    3 | async function foo(): Awaitable<void> {
    4 |   while(true) {
[1] 5 |     await bar();
    6 |   }
    7 | }
```

However, because it does not have an 'interesting' extension, `hh_client` (without `--lint`) should ignore it.

Atry:
```
$ hh_client --lint tests/examples/HHClientLinter/invalid_null_check.hack.in 
No lint errors!
$ mv tests/examples/HHClientLinter/invalid_null_check.hack.in   tests/examples/HHClientLinter/invalid_null_check.hack
$ hh_client --lint tests/examples/HHClientLinter/invalid_null_check.hack
Lint[5611] Invalid null check: This expression will always return false.
A value of type int can never be null. [1]

tests/examples/HHClientLinter/invalid_null_check.hack:14:7
    12 | function invalid_null_check(): void {
    13 |   $cannot_be_null = 42;
[1] 14 |   if ($cannot_be_null is null) {
    15 |     throw new Exception();
    16 |   }

```

Atry: My understanding is that `Lint[5611]` requires some type information only available for loaded files, while `Lint[5583]` applies to both loaded files and unloaded files.

Atry:
Close this PR because ignoring the temporary files would implicitly disable `Lint[5611]`

fredemmott:
nice catch :)

PR #434, opened 2022-01-20
Update hhvm-autoload to 3.3.1 to support nightly HHVM

Atry:


fredemmott (approved):

fredemmott:
This appears fine - but why is it needed? Composer should automatically use the latest available version

PR #439, opened 2022-02-28
Add linter to identify unreachable code

ryangreenberg:
This PR adds UnreachableCodeLinter which flags any `return`, `throw`, or `continue` statement that is followed by other code at the same scope (CompoundStatement is used to approximate "same scope").

This is probably best implemented in the typechecker which can do better analysis than a linter, but as an intermediate solution this might help folks identify potential bugs.

### DIFF
--- a/src/Linters/UnreachableCodeLinter.hack
+++ b/src/Linters/UnreachableCodeLinter.hack
@@ -35,13 +35,18 @@ final class UnreachableCodeLinter extends ASTLinter {
         $statements = $parent->getStatements();
         invariant($statements is nonnull, 'parent list of stmt cannot be null');
         $children = $statements->getChildren();
-        $return_idx = C\find_key($children, $c ==> $c === $stmt) ?? null;
-        invariant(
-            $return_idx is nonnull,
-            'stmt must be a child of parent list',
-        );
+        $stmt_idx = C\find_key($children, $c ==> $c === $stmt) ?? null;
 
-        if ($return_idx < C\count($children) - 1) {
+        // if the statement is not a direct child of the CompoundStatement,
+        // then it occurs in a position where it cannot cause unreachable code
+        //
+        // For example:
+        // if ($cond) return;
+        if ($stmt_idx is null) {
+            return null;
+        }
+
+        if ($stmt_idx < C\count($children) - 1) {
             if ($stmt is ThrowStatement) {
                 $op = 'throw';
             } else if ($stmt is ReturnStatement) {
--- a/src/Linters/UnreachableCodeLinter.hack
+++ b/src/Linters/UnreachableCodeLinter.hack
@@ -12,50 +12,60 @@ namespace Facebook\HHAST;
 use namespace HH\Lib\{C, Str};
 
 final class UnreachableCodeLinter extends ASTLinter {
-    const type TConfig = shape();
-    const type TNode = IStatement;
-    const type TContext = CompoundStatement;
+  const type TConfig = shape();
+  const type TNode = IStatement;
+  const type TContext = CompoundStatement;
 
-    <<__Override>>
-    public function getLintErrorForNode(
-        CompoundStatement $parent,
-        IStatement $stmt,
-    ): ?ASTLintError {
-        // Evaluate all IStatement because we need to look for additional statements
-        // after return, throw, continue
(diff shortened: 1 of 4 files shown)


src/Linters/UnreachableCodeLinter.hack
```diff
@@ -0,0 +1,67 @@
...
+
+        // if the statement is not a direct child of the CompoundStatement,
+        // then it occurs in a position where it cannot cause unreachable code
+        //
+        // For example:
+        // if ($cond) return;
+        if ($stmt_idx is null) {
+            return null;
+        }
+
```
lexidor, line 48:
```suggestion
        $children = $statements->getChildren();

        // if the statement is not a direct child of the CompoundStatement
        // or it is the last statement of the CompoundStatement,
        // then it occurs in a position where it cannot cause unreachable code
        //
        // For example:
        // if ($cond) return;
        // if ($cond) { $a = 5; return "last statement"; }
        if (!C\contains($children, $stmt) || C\last($children) === $stmt) {
            return null;
        }

```

src/Linters/UnreachableCodeLinter.hack
```diff
@@ -0,0 +1,67 @@
...
+            return new ASTLintError(
+                $this,
+                Str\format('This %s statement creates unreachable code', $op),
+                $stmt,
+            );
+        } else {
+            return null;
+        }
+    }
```
lexidor, line 66:
```suggestion
        return new ASTLintError(
            $this,
            Str\format(
                'This %s statement creates unreachable code',
                $stmt->getKeyword()->getText(),
            ),
            $stmt,
        );
```

lexidor (approved):
Thank you for writing this linter. Having HHAST check for this will be a great addition to the "this-is-likely-an-error" set of linters.

I have made some small structural suggestions. I also noticed that you have indented the files with four spaces, rather than two.

:+1:

ryangreenberg:
Thanks for the suggestions, @lexidor. I'll rework the structure with your suggestions and update here. Not sure what happened with the indenting; maybe I have vscode configured in an unexpected way. I'll run hackfmt directly on the affected files.

tests/UnreachableCodeLinterTest.hack
```diff
@@ -1,3 +1,6 @@
+
```
lexidor, line 1: Hackfmt created some leading whitespace here.

src/Linters/UnreachableCodeLinter.hack
```diff
@@ -12,50 +12,60 @@ namespace Facebook\HHAST;
...
+    //     if ($cond) { $a = 5; return "last statement"; }
+    //     if ($cond) return;
+    if (C\last($children) === $stmt || !C\contains($children, $stmt)) {
+      return null;
     }
+
+    if ($stmt is ThrowStatement) {
+      $op = 'throw';
+    } else if ($stmt is ReturnStatement) {
+      $op = 'return';
```
lexidor, line 60: This if-else chain is equivalent to `$stmt->getKeyword()->getText()`.

lexidor:
> For the hackfmt mismatches in this PR, I don't see any `hackfmt` settings defined in the [repo's .hhconfig](https://github.com/hhvm/hhast/blob/cade7dc28053b2731cfc0b54de8773771e4c04c6/.hhconfig). Maybe we want to define something there so contributions are autoformatted with the repo's preference?

As far as I understand it, vscode setting prevail over hhconfig settings when there is a conflict. So we'd need to add a `.vscode/setting.json` file and commit that too.

fredemmott:
In general, I'd rather avoid duplicating defaults (and overriding them if they change) in tool-specific configuration files. We mostly want hhast to stick with whatever hackfmt does by default, even if that changes in the future.

Are your existing vscode settings for non-Hack projects? Can they be scoped to ignore hack projects?

if they're for another hack project, can the preferences for that project be moved into its' .hhconfig instead of a vscode-specific file?

ryangreenberg:
I fixed the failing lint test and simplified the branch to get the statement keyword (thanks @lexidor).

WRT formatting, I checked my editor setting and I suspect the output I got was related to having tab stops set to 4 and detect indentation disabled (I don't think I have any Hack/hackfmt specific settings configured in the editor). Makes sense not to restate the hackfmt defaults in the repo's config.

PR #444, opened 2022-04-08
Update codegen for 4.156 and require 4.156

lexidor:
This version of hhast supports hhvm 4.156.
The `$obj->get#Version()` syntax was removed from hhvm.
We therefore can't parse older code correctly, since it might
use this enum label call syntax.
Fixes crash when using ext_watchman.
Watchman does not like usage of a constant that was just codegenned.

### DIFF
--- a/tests/NamespacePrivateLinterTest.hack
+++ b/tests/NamespacePrivateLinterTest.hack
@@ -20,7 +20,7 @@ final class NamespacePrivateLinterTest extends TestCase {
   <<__Override>>
   public function getCleanExamples(): vec<(string)> {
     return vec[
-      tuple('<?hh function test();'),
+      tuple('<?hh function test() {}'),
     ];
   }
 }
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,6 +5,7 @@ src/codegen/data/ export-ignore
 tests/ export-ignore
 codegen/**/* linguist-generated
 .hhconfig export-ignore
+.hhvmconfig.hdf export-ignore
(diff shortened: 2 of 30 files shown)


fredemmott:
Thanks! Can you split out the autoloader configuration to a separate commit from the update please?

tests/NamespacePrivateLinterTest.hack
```diff
@@ -20,7 +20,7 @@ final class NamespacePrivateLinterTest extends TestCase {
   <<__Override>>
   public function getCleanExamples(): vec<(string)> {
     return vec[
-      tuple('<?hh function test();'),
+      tuple('<?hh function test() {}'),
```
fredemmott, line 23: This needs to be kept for docs.hhvm.com to be able to analyze builtins - this is sort-of valid when `<<__Native>>` is present

lexidor: Oh, using `;` as the body of a function was banned. There are no code examples left that use it. The codegen changed its type. Maybe a manual syntax example would restore the codegen to `CompoundStatement | SemiColon`?

fredemmott: Yep, I'm adding a manual example which should result in that

It's banned in 'real' hack code, but `<<__Native>>` is an exception, and `<<__Native>>` is only valid in code that is built in to HHVM or an extension

fredemmott:
Replaced by #445 - thanks, this was really useful, especially for spotting the enumlabel issue!

PR #448, opened 2022-04-12
Update hhast schema to reflect changes in facebook/hhvm

alexeyt:


codegen/syntax/TypeConstDeclaration.hack
```diff
@@ -20,7 +20,7 @@ final class TypeConstDeclaration extends Node implements IClassBodyDeclaration {
   private TypeToken $_type_keyword;
   private NameToken $_name;
   private ?Node $_type_parameters;
-  private ?TypeConstraint $_type_constraint;
+  private ?NodeList<TypeConstraint> $_type_constraints;
```
fredemmott, line 23: As dicussed, this is a breaking change, as it can break existing callers

fredemmott (approved):

PR #449, opened 2022-04-12
Update hhvm dependency

alexeyt:


fredemmott (approved):
Expecting nightlies to pass, everything else to fail until HHVM 4.157 is out

PR #452, opened 2022-04-19
Improve the commands to update schema

Atry:
- Use `code` instead of `edit` given that the project has been opened in VS Code
- Add `-y` to apt commands.


docs/updating-schema.md
```diff
@@ -1,10 +1,9 @@
 HHAST is closely coupled to Hack's syntax tree, so if the schema of `hh_parse`
 changes, HHAST needs updating.
 
-0. Get a github codespace or use vscode devccontainers. If a codespace, pick the biggest one you can for this (in general, the smallest are good enough, but here more power is useful).
-1. apt update; apt install TEXT_EDITOR_OF_YOUR_CHOICE
-2. edit /etc/apt/sources.list - remove `-4.x` suffix from dl.hhvm.com repository
-3. apt update; apt remove hhvm; apt install hhvm-nightly (or `apt install hhvm` for latest stable release)
+1. Get a github codespace or use vscode devcontainers. If a codespace, pick the biggest one you can for this (in general, the smallest are good enough, but here more power is useful).
+2. `code /etc/apt/sources.list` - remove `-4.x` suffix from dl.hhvm.com repository
```
alexeyt, line 5: This change assumes people always use the vscode editor? IIUC Fred did not, and I certainly do not :-) I am OK with your chanbes to step 3 (adding -y), and I am OK removing step 1 (many editors are already available, if you want a non-default one it's up to you to install it), but I would like this step to remain as "edit /etc/apt/sources.list"

alexeyt (changes requested):

PR #454, opened 2022-04-25
Update AST schema to 2022-04-20-0001

Atry:


src/__Private/is_compatible_schema_version.hack
```diff
@@ -21,14 +21,15 @@ use const Facebook\HHAST\SCHEMA_VERSION;
...
-    SCHEMA_VERSION === '2022-04-13-0002',
+    SCHEMA_VERSION === '2022-04-20-0001',
     '%s needs updating',
     __FILE__,
   );
-
-  if ($other_version === SCHEMA_VERSION) {
-    return true;
+  switch ($other_version) {
+    case '2022-04-20-0001': // Adding enum modifiers and alias modifiers
```
alexeyt, line 29: Should this one be `case SCHEMA_VERSION:`? Not sure it makes a huge difference, but it feels weird to not use the constant in this logic.

alexeyt (approved):

src/__Private/is_compatible_schema_version.hack
Atry: If we do `case SCHEMA_VERSION: // Adding enum modifiers and alias modifiers`, the comment seems weird. It reads as "the current schema version adds enum modifiers and alias modifiers", which does not sound like an eternal truth.

PR #459, opened 2022-05-03
Add UnusedPipeVariableLinter

ryangreenberg:
This PR introduces `UnusedPipeVariableLinter` to identify bugs where the pipe variable `$$` is missing from a pipe operation.

I recently encountered a bug in a pipe operation where the right-hand side re-used the initial variable, discarding the intended input in `$$`:

```hack
$output = htmlspecialchars($html) |> format($html);
```

Note that the docs for ["Expressions And Operators: Pipe"](https://docs.hhvm.com/hack/expressions-and-operators/pipe) say:

> The binary pipe operator, `|>`, evaluates the result of a left-hand expression and stores the result in `$$`, the pre-defined pipe variable. The right-hand expression _must_ contain at least one occurrence of `$$`.

If it's a bug that the typechecker does not produce an error for a missing `$$`, we can close this PR and file an issue in https://github.com/facebook/hhvm.

### DIFF
--- a/src/Linters/UnusedPipeVariableLinter.hack
+++ b/src/Linters/UnusedPipeVariableLinter.hack
@@ -31,7 +31,7 @@ final class UnusedPipeVariableLinter extends ASTLinter {
             return new ASTLintError(
                 $this,
                 "Missing pipe variable in right-hand side of pipe operator",
-                $rhs,
+                $expr,
             );
         }
 
--- a/src/Linters/UnusedPipeVariableLinter.hack
+++ b/src/Linters/UnusedPipeVariableLinter.hack
@@ -16,7 +16,7 @@ final class UnusedPipeVariableLinter extends ASTLinter {
 
   <<__Override>>
   public function getLintErrorForNode(
-    Script $script,
+    Script $_script,
     BinaryExpression $expr,
   ): ?ASTLintError {
     $op = $expr->getOperator();
--- a/src/Linters/UnusedPipeVariableLinter.hack
+++ b/src/Linters/UnusedPipeVariableLinter.hack
@@ -10,31 +10,30 @@
 namespace Facebook\HHAST;
 
 final class UnusedPipeVariableLinter extends ASTLinter {
-    const type TConfig = shape();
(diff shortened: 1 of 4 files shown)


lexidor:
> If it's a bug that the typechecker does not produce an error for a missing $$, we can close this PR and file an issue in https://github.com/facebook/hhvm.

https://github.com/facebook/hhvm/issues/8698

There is already an issue tracking this 😉.
Please don't let this convince you to withdraw this linter. Pipes without $$ are fishy it should be flagged. A lint rule could be deployed far quicker than a language change. I would be surprised if www would typecheck after 2.5 years since the unintentional (?) removal of the check. Strongly in favor of shipping this linter, even if it becomes a language level error later.

ryangreenberg:
Two of the test suites are failing because they could not run hhvm and two are failing with this error:

```
 Server launched with the following command:
  	'/usr/local/Cellar/hhvm-nightly/2022.05.03/bin/hh_server' '-d' '/Users/runner/work/hhast/hhast' '--waiting-client' '8'
  Running in daemon mode
  Server connection took over 2.0 seconds. Refreshing...
  Typing[4297] Was expecting an array or collection but type is unknown [1]
  -> It is unknown because the type of the lambda parameter could not be determined. Please add a type hint to the parameter [2]
  
  src/__Private/XHProf.hack:32:53
        30 |     invariant(self::$enabled === true, "Can't disable twice");
        31 |     self::$enabled = false;
  [1,2] 32 |     $raw = Dict\map(\xhprof_disable(), $v ==> (int) $v['wt']);
        33 | 
        34 |     $inclusive = dict[];
  
  1 error found.
  Error: Process completed with exit code 2.
```

This seems like an unrelated

lexidor:
Yes, I ran into that too. You can add a `dict<arraykey, mixed>` before `$v` and explicitly reformat. This makes the tests pass again without conflict with my branches.

Atry:
Retrigger the CI

ryangreenberg:
Retriggered CI again to get the build to pass on macOS. I think this PR is now ready for review/discussion.

PR #460, opened 2022-05-05
Add explicit type to lambda argument.

alexeyt:
This makes it typecheck with nightly hh_client.


Atry (approved):

Atry:
I think Homebrew bottles need to be rebuilt

PR #506, opened 2022-08-11
Fix some shortcomings of UnusedParameterLinter

lexidor:
- Variadic parameters were not checked
   function _(int ... $x): void {} now emits an error for $x.
- Lambda parameters are now linted in unparenthesized params.
   $x ==> 0 now emits an error for $x.
   Same applies for (int ... $x) ==> 0.
 - Added explicit tests for old style variadics.
   The linter handled them fine, but there was no test for it.
Checking for unparenthesized lambda parameters required a new linter.
The UnusedParameterLinter checked against ParameterDeclaration.
$x ==> 0; does not introduce one, so a new TNode was required.

This PR may cause old HHAST_FIXME comments to break.
You will have to upgrade:
/\*HHAST_FIXME[UnusedParameter]\*/ ($x) ==> 0; to
/\*HHAST_FIXME[UnusedLambdaParameter]\*/ ($x) ==> 0;
I suspect this linter is rarely suppressed, since the autofix is easier.

### DIFF
--- a/src/Linters/UnusedLambdaParameterLinter.hack
+++ b/src/Linters/UnusedLambdaParameterLinter.hack
@@ -1,0 +1,69 @@
+/*
+ *  Copyright (c) 2017-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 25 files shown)


Atry:
Looks good to me.

PR #517, opened 2022-09-17
Add an optional lintMarkerAllowList

### LINKED ISSUE #458: [ RFC ] Adding an optional allow list to hhast-lint.json
_The following describes a suggestion. This issue is not meant to be implemented right away. This might not be a direction hhast wants to go in._

---

I would like to propose an allow_list for HHAST_FIXME comments, analogous to the hhconfig setting `allowed_fixme_codes_strict` introduced in hhvm 4.62. If a FIXME is found for a linter that is not in the allow list, an error like the following would be emitted. _You may not use a comment to suppress a {{LintErrorName}} error. See lintFixmeAllowList in hhast-lint.json._ The default value (unspecified) should be equivalent to "allow all fixmes" rather than "allow none".

Some linters should always be adhered to and adding a FIXME for them is always a bad idea. Other linters have valid uses for a FIXME. This use of await-in-a-loop is critical to the functionality.
```HACK
function wait_until_service_is_ready_async(string $url, Deadline $deadline): Awaitable<bool> {
  while($deadline->hasTimeLeft()) {
    // HHAST_IGNORE_ERROR[DontAwaitInALoop]
    $status = await get_http_status_async($url);
    switch ($status) {
      case 200:
        return true;
      case 404:
        throw new RuntimeException(Str\format('Service %s does not exist', $url);
      case 0:
      case 503:
        $sleep = $deadline->microsLeftOr(25000) - 5000;
        if ($sleep <= 0) {
          return false;
        }
        //HHAST_IGNORE_ERROR[DontAwaitInALoop]
        await Asio\usleep($sleep);
      default:
        // etc...
    }
  }
  return false;
}
```

A configuration file for this project could look like:
```JSON
{
  "roots": ["src/"],
  "lintFixmeAllowList": [
    "Facebook\\HHAST\\DontAwaitInALoopLinter",
  ],
  "overrides": [
    {
      "patterns": ["src/codegen/**"],
      "lintFixmeAllowList": [
        "YourCorp\\SomeLinter"
      ]
    }
  ]
}
```

lexidor:
fixes #458


src/Linters/HHClientLinter.hack
```diff
@@ -18,8 +18,13 @@ final class HHClientLinter implements Linter {
   use LinterTrait;
   use SuppressibleTrait;
 
-  const type TConfig =
-    shape(?'ignore_except' => vec<int>, ?'ignore' => vec<int>, ...);
```
lexidor, line 22: I have added `ignoreExcept` as an alternative spelling of `ignore_except`.
I wanted to match the case of `lintMarkerAllowList` in global scope.

The following looks wrong:
```JSON
{
  "ignore_except": [],
  "lintMarkerAllowList": []
}
```

Atry:
Close and reopen to trigger CI

lexidor:
Asking for feedback for HHClientLinter.
This linter is special, because it is an amalgamation of hundreds of rules.
The option "allow comments for HHClientLinter" could not be a simple binary toggle.
I have added the lintMarkerAllowList option to HHClientLinter to allow for granular tuning.

I completely ignore the global lintMarkerAllowList in HHClientLinter. So when `"lintMarkerAllowList": ["UnusedUseClause"]` is configured HHClientLinter will still honor its own suppression comments.

I did this, because of the following confusing example.

```JSON
{
  "roots": ["src/"],
  "lintMarkerAllowList": [],
  "linterConfigs": {
    "HHClientLinter": {
      "lintMarkerAllowList": [5555]
    }
  }
}
```

This is conflicting, because the global list says, no markers allowed, but HHClientLinter allows marker 5555. I would expect HHClientLinter to respect its own config here. But this means, if you want to disallow all markers, you need to set two empty lists. One for HHClientLinter and one at the global level.

I also considered to error out when HHClientLinter receives a lintMarkerAllowList in its own config when shouldAllowSuppressionComments() is false. But this requires you to repeat yourself. First add HHClientLinter linter to the global list and then fine-tune the HHClientLinter config.

Which of these two would be right for the hhast project? Which would be the least surprising?

Atry:
What is the difference between the new settings and existing `ignore` and `ignore_except` settings for `HHClientLinter`?

Atry:
Do we really need the `lintMarkerAllowList` setting for `HHClientLinter` if there is such the `lintMarkerAllowList` global setting?

lexidor:
> Do we really need the `lintMarkerAllowList` setting for `HHClientLinter` if there is such the `lintMarkerAllowList` global setting?

The global list contains only classnames. HHClientLinter has many sublinters:
 - SketchyNullCheckLinter
 - BooleanExpressionAlwaysFalseLinter
 - DontUseTheCloneOperatorLinter
 - DontUseSafeMemberAccessOnNonnullLinter
 - DontImplicitlyConvertStringToBoolLinter
These are not classnames. The sublinters of HHClientLinter are unnamed. They can't be included in the global list.

Atry:
There is no "sublinter" concept in HHAST. A marker is for a lint rule, whose name could be either a class name or an integer code. Given the setting name `lintMarkerAllowList`, I would expect it contains both class names and integer codes.

Atry:
Given that the proposed setting is for markers, not linters, it would be straightforward to evaluate the allow list when parsing markers globally, instead of implementing in each linter.

- Parsing line based marker at https://github.com/hhvm/hhast/blob/5bc52f7af5df17b65a513227483e56d75fa7cafc/src/File.hack#L42
- Parsing AST based marker at https://github.com/hhvm/hhast/blob/5bc52f7af5df17b65a513227483e56d75fa7cafc/src/Linters/suppress_ast_linter_error.hack#L19

lexidor:
> There is no "sublinter" concept in HHAST. A marker is for a lint error code, whose name could be either a class name or an integer code. Given the setting name `lintMarkerAllowList`, I would expect it contains both class names and integer codes.

That would be worth it. I'll type it as `keyset<arraykey>`.

Atry:
`arraykey` might be better. I typed it as `string` because it was `string` long before we introduce `HHClientLinter` and the integer error code and changing it would break existing code.

lexidor:
> Given that the proposed setting is for markers, not linters, it would be straightforward to evaluate the allow list when parsing markers globally, instead of implementing the feature in each linter.
> 
>     * Parsing line based markers at https://github.com/hhvm/hhast/blob/5bc52f7af5df17b65a513227483e56d75fa7cafc/src/File.hack#L42
> 
>     * Parsing AST based markers at https://github.com/hhvm/hhast/blob/5bc52f7af5df17b65a513227483e56d75fa7cafc/src/Linters/suppress_ast_linter_error.hack#L19

I am assuming you are proposing to add a new function besides is_linter_error_suppressed that returns a tri-state.

```HACK
enum SuppressionState : int {
  NOT_SUPPRESSED = 0;
  SUPPRESSED = 1;
  DISALLOWED_SUPPRESSION = 2;
}
```

I could then reimplement the is_linter_error_suppressed as `get_suppression_state($linter, $node, $root, /*is suppression allowed*/ true) === SuppressionState::SUPPRESSED`.

I am not so sure about how I would change the signature of erroCodesToSuppress. Any suggestions?

src/Linters/HHClientLinter.hack
Atry: The internal coding convention in Meta Platforms prefers lowercase_with_underscores for shape keys, although existing HHAST code base includes both lowerCamelCases and lowercase_with_underscores for shape keys.

https://github.com/hhvm/hhast/blob/d0a746ad3aabdfb053e7505aaf09583a8e813284/src/__Private/Resolution/get_uses_directly_in_scope.hack#L88

https://github.com/hhvm/hhast/blob/d0a746ad3aabdfb053e7505aaf09583a8e813284/src/Linters/DontDiscardNewExpressionsLinter.hack#L15

Atry:
I would let `is_linter_error_suppressed` and `errorCodesToSuppress` simply ignore markers no in the allow list, and create a new linter to warn markers no in the allow list.

src/Linters/HHClientLinter.hack
lexidor: I usually prefer snake_case as well, but I am trying to stay consistent with the hhast-lint.json format, which is 95% camelCase and 5% English with spaces.

### camelCase

`builtinLinters`, `extraLinters`, `disabledLinters`, `disabledAutoFixes`, `disableAllAutoFixes`, `disableAllLinters`, `linterConfigs`, `namespaceAliases`, `exceptionSuffixes`.

### English with spaces

`line ending`

The latter is a mistake on my part, but this can be fixed, since the linter doesn't observe the value for `"line ending"` (yet). So I can safely migrate this over to `lineEnding` without breaking BC.

lexidor:
> I would create a new linter to raise lint errors for markers not in the allow list, and let all other linters simply ignore markers not in the allow list.

That is a very interesting approach I had not considered before. That is a great idea! Thanks for your input. I will rework my PR.

If that new linter would be impossible to suppress, we'd already be done. That could be done by automatically adding `DontUseSuppressionMarkersThatAreNotDeclaredInLintMarkerAllowListLinter` to the lintMarkerAllowList, because that would cause and infinite suppression regression.

Then we barely have to change anything about suppression comment logic. We just need to defang `HHAST_IGNORE_ALL` for DontUseSuppressionMarkersThatAreNotDeclaredInLintMarkerAllowList.


```HACK
// We just need to defang this IGNORE_ALL, since this would suppress itself.
/*HHAST_IGNORE_ALL[DontUseSuppressionMarkersThatAreNotDeclaredInLintMarkerAllowList]*/

// After that...
/*HHAST_FIXME[DontUseSuppressionMarkersThatAreNotDeclaredInLintMarkerAllowList]
  Oh wait, I can't, since it is turtles all the way down. */
/*HHAST_FIXME[DontUseSuppressionMarkersThatAreNotDeclaredInLintMarkerAllowList]
  Suppressing the don't use markers linter*/
/*HHAST_FIXME[DontUseSuppressionMarkersThatAreNotDeclaredInLintMarkerAllowList]
  Suppressing the don't suppress MustUseBracesForControlFlow linter */
/*HHAST_FIXME[MustUseBracesForControlFlow]*/
if ($xyzzy) return;
```

src/Linters/HHClientLinter.hack
Atry: Fair enough. Let's document `hhast-lint.json` and `TConfig` about the state of lowerCamelCases usage.

lexidor:
Converting to draft, will redo using a new linter. `DontUseSuppressionMarkersThatAreNotDeclaredInLintMarkerAllowListLinter` is a placeholder name btw.

PR #537, opened 2023-01-23
use ffp_parse_string instead of shelling out to hh_parse

ssandler:
This yielded a *dramatic* speedup in HHAST for us, and as far as I can tell is absolutely equivalent.

Since the built-in type of `ParserCache` is opaque, I had to `as dict<_, _>` it, and to avoid adding FIXMEs I changed the type annotations where it's used to allow arraykey keys.

The built-in parser doesn't seem to support old style xhp children declarations, but since those have been deprecated for several years I think it's ok to remove support for them in HHAST and the associated migration, so I've done that as well.


ssandler:
@lexidor good point! I'm incorporating that and also working on fixing CI.

ssandler:
CI is broken for other reasons, moved the work to get that working into https://github.com/hhvm/hhast/pull/538

fredemmott (approved):

PR #538, opened 2023-01-23
Ssandler get ci working

ssandler:
HHAST CI has been broken for over a month due to there not being a build on latest ubuntu, and also due to a schema change in the AST which broke parsing xhp children.

This PR updates the CI to **only** run on the latest HHVM LTS (4.168) and on Ubuntu 20.04, where there is a working build. It also updates the schema to work with that release, and updates `composer.json` required HHVM version.


.github/workflows/build-and-test.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: hhvm/actions/hack-lint-test@master
         with:
-          hhvm: nightly
+          hhvm: 4.168
```
fredemmott, line 50: This effectively disables the job and HHAST auto-updating for nightlies - that might be fine (@Wilfred ? ), but if so, it's probably better to remove the job rather than change this

ssandler: Makes sense. It's currently broken on nightly, but for different reasons. Some other breaking change upstream causes it to fail https://github.com/hhvm/hhast/actions/runs/3991497523/jobs/6846355294

I don't think we can keep up with getting this to work for nightlies so might be ok to just go back to manual schema updates for the foreseeable future. Could remove or comment out this job.

ssandler: I removed the job and also reverted the docs PR about it back to the old instructions, which I followed to get the codegen working.

PR #544, opened 2023-05-24
Banish hh ignore error from codegen

lexidor:
Fixes:
 - No more `HH_IGNORE_ERROR` directives in codegen.
 - No more `HH_IGNORE_ERROR` directives in hand rolled nodes.
 - Various `HH_IGNORE_ERROR` -> `HH\FIXME\UNSAFE_CAST<Ta, Tb>(...)` migrations.
   - It is not that I particularly like UNSAFE_CAST, but they are better than a 4110 fixme, right?
 - Remove `HH_IGNORE_ERROR` from declarations.

### DIFF
--- a/src/__Private/codegen/CodegenSyntax.hack
+++ b/src/__Private/codegen/CodegenSyntax.hack
@@ -760,7 +760,8 @@ final class CodegenSyntax extends CodegenBase {
     $is_a_known_type = $is_node_listy && $rest === '<Node>' ||
       $is_list_itemy && $rest === '<?Node>';
 
-    return
-      $is_a_known_type ? $expression.' as '.$open_generic : $best_effort_fixme;
+    return $upper_bound === $expected_type
+      ? $expression.' as '.$open_generic
+      : $best_effort_fixme;
   }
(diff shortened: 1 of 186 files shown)


src/__Private/codegen/CodegenSyntax.hack
```diff
@@ -706,4 +707,60 @@ final class CodegenSyntax extends CodegenBase {
...
+
+    $best_effort_fixme = Str\format(
+      '\\HH\\FIXME\\UNSAFE_CAST<%s, %s>(%s as %s, \'Open for sound approaches that are not O(n).\')',
+      $upper_bound,
+      $expected_type,
+      $expression,
+      $open_generic,
+    );
+
+    $is_a_known_type = $is_node_listy && $rest === '<Node>' ||
```
lexidor, line 760: `$upper_bound === $expected_type`

lexidor:
CI doesn't run here anymore, so you'll have to look at the [CI of my clone](https://github.com/lexidor/hhast/actions/runs/7314903128).

